PR #4, opened 2020-04-18
Add charts 

KyleMcMaster:
Add two charts to payroll tab 

- one to show employee wages distribution
- one to show department wages distribution

removing (disabling) an employee in the employees list will affect the payroll tabs and distribution of the charts (try disabling NTG).

Interested in feedback on how to refactor the `payroll-charts.component.ts`'s data processing methods


KyleMcMaster:
@all-contributors please add @eyev for review and code

payroll-processor-client/src/app/payroll/payroll-charts/payroll-charts.component.ts
```diff
@@ -0,0 +1,83 @@
...
+  };
+
+  constructor(private dataService: DataService) {
+    this.prepData();
+  }
+
+  ngOnInit(): void {}
+
+  prepData() {
+    this.prepDepartmentDistributionData();
```
seangwright, line 37: I'd keep these in the constructor.

If you find yourself wanting to break up a large function/method into smaller ones and the bigger one then calls all the smaller ones, there's probably a 'single responsibility' violation.

ex)

```typescript
someBigMethod() {
   // do thing a
   
   // do thing b

   // ...

   // do thing z
}

// becomes

someBigMethod() {
   this.doThingA();
   this.doThingB();

   // ...

   this.doThingZ();
}
```

Encapsulating these procedural operations hides the complexity, it doesn't manage or resolve it.

payroll-processor-client/src/app/payroll/payroll-charts/payroll-charts.component.ts
```diff
@@ -0,0 +1,83 @@
...
+    this.prepEmployeeDistributionData();
+  }
+
+  prepEmployeeDistributionData() {
+    const payroll = this.dataService.getPayroll();
+    const employees = this.dataService.getEmployees();
+    const map: Map<string, number> = new Map();
+
+    payroll.forEach((p) => {
```
seangwright, line 46: As a convention, avoid `.forEach()`.

If you need an imperative loop, use `for (const val of values) { ... }`

payroll-processor-client/src/app/payroll/payroll-charts/payroll-charts.component.ts
```diff
@@ -0,0 +1,83 @@
...
+      if (p.employeeStatus !== 'ACTIVE') {
+        return;
+      }
+      return map.has(p.employeeName)
```
seangwright, line 50: Returning a value in a `.forEach()` doesn't have any meaning. You'd want a `.map()`, `.filter()`, or `.reduce()` if you intend to return a value.

payroll-processor-client/src/app/payroll/payroll-charts/payroll-charts.component.ts
```diff
@@ -0,0 +1,83 @@
...
+        : map.set(p.employeeName, p.grossPayroll);
+    });
+
+    this.employeeDistributionData = Array.from(
+      map,
+      ([key, value]) => new NameValue(key, value),
+    );
+  }
+
+  prepDepartmentDistributionData() {
```
seangwright, line 61: Without getting into too much architecture (rxjs, akita, ect...), we could at least move these data processing operations out of the component and into a "Presenter".

Think of a Presenter as a service class that is directly tied to this component.

Components are glue between state and HTML, any other processing should be put somewhere else if possible.

To make this change you can create a new service class `payroll-charts.presenter.ts` and inject it into this component.

Give it 2 methods `prepDepartmentDistributionData`, and `prepEmployeeDistributionData` that return your `Map`s.

Call those methods in this component's constructor to assign to the component `Map` fields.

payroll-processor-functions/src/payroll-processor-functions/Features/Employee/EmployeesGetTrigger.cs
```diff
@@ -10,9 +10,9 @@ namespace PayrollProcessor.Functions.Features.Employee
 {
     public class EmployeesGetTrigger
     {
-        [FunctionName("EmployeesGetTrigger")]
+        [FunctionName(nameof(EmployeesGetTrigger))]
```
seangwright, line 13: I like this idea - I think in my PR I should use `nameof()` and pass the function method name (in this case it would be `nameof(Run)`)

payroll-processor.code-workspace
```diff
@@ -10,7 +10,7 @@
     },
     {
       "path": "payroll-processor-functions",
-      "name": "funtions"
```
seangwright, line 13: I think I made this typo 100 times in the PPT app.

payroll-processor-client/src/app/payroll/payroll-charts/payroll-charts.component.ts
```diff
@@ -0,0 +1,83 @@
...
+import { DataService } from 'src/app/data/data-service';
+import { NameValue } from 'src/app/data/state/data-model';
+
+@Component({
+  selector: 'app-payroll-charts',
+  templateUrl: './payroll-charts.component.html',
+  styleUrls: ['./payroll-charts.component.scss'],
+})
+export class PayrollChartsComponent implements OnInit {
+  departmentDistributionData: NameValue[] = [];
```
seangwright, line 15: When you make the changes to add the presenter service, mark these as `readonly` since you'll be assigning the values in this component's constructor.

payroll-processor-client/src/app/payroll/payroll-charts/payroll-charts.component.ts
```diff
@@ -0,0 +1,83 @@
...
+  employeeDistributionData: NameValue[] = [];
+  view: [number, number] = [700, 400];
```
seangwright, line 17: All these chart configuration values can be `readonly` as well.

payroll-processor-client/src/app/data/state/data-model.ts
```diff
@@ -0,0 +1,9 @@
+export class NameValue {
```
seangwright, line 1: I'd recommend using classes when you need to generate derived state:

```typescript
class User {
   firstName = 'test';
   lastName = 'user';

   get fullName() { return `${this.firstName} ${this.lastName}`; }
}
```

If you don't have derived state, an interface is a better option because it's committing to less 'structure' too early.

```typescript
interface NameValue {
  readonly name: string;
  readonly value: number;
}
```

payroll-processor-client/src/app/payroll/payroll-charts/payroll-charts.component.html
```diff
@@ -0,0 +1,32 @@
+<div class="container">
+  <div class="row">
+    <div class="col-6">
+      <h6>Employee Payroll Distribution</h6>
+      <ngx-charts-pie-chart
+        [view]="view"
```
seangwright, line 6: There's not much getting around these large list of properties for highly configurable components.

There are ways around it, but they're not worth the abstraction atm.

KyleMcMaster:
Abandon/Archive this feature until use cases are documented for future enhancements

PR #5, opened 2020-04-21
Add vscode workspace / api launch/tasks

seangwright:
Adds a VS Code workspace file and config files for the API for launching/debugging.

Opening the repo in VS Code will prompt the dev to "open the workspace" which presents the repo more like a VS solution.

We can start adding settings/config at the workspace level now.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -378,9 +378,6 @@ pubspec.lock
 !.devcontainer/recommended-devcontainer.json
 !.devcontainer/recommended-Dockerfile
 .settings/
-.vscode/launch.json
-.vscode/settings.json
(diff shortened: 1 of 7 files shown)


.gitignore
```diff
@@ -378,9 +378,6 @@ pubspec.lock
 !.devcontainer/recommended-devcontainer.json
 !.devcontainer/recommended-Dockerfile
 .settings/
-.vscode/launch.json
```
seangwright, line 381: Anything in `.vscode/` should be included in the repo for shared project configuration.

KyleMcMaster: Good catch!

PR #6, opened 2020-04-21
Feat/client env service

seangwright:
Adds a new `EnvService` abstraction to prevent the app from consuming `environment.ts` directly and improve testability.

Also adds an `Environment` abstraction to force all environment files to match the same schema and give a consistent shape to environment consumption in the app.

Also adds some tsconfig updates that we can slowly uncomment to improve the type checking.

### DIFF
--- a/payroll-processor-client/src/app/data/data-service.ts
+++ b/payroll-processor-client/src/app/data/data-service.ts
@@ -4,6 +4,7 @@ import { Employee } from './state/employee-model';
 import { catchError } from 'rxjs/operators';
 import { throwError } from 'rxjs';
 import { Payroll } from './state/payroll-model';
+import { EnvService } from '../shared/env.service';
 
 @Injectable({
   providedIn: 'root',
@@ -12,16 +13,17 @@ export class DataService {
   private employees: Employee[] = [];
   private payroll: Payroll[] = [];
 
-  constructor(private http: HttpClient) {
+  private readonly apiUrl: string;
+
+  constructor(private http: HttpClient, envService: EnvService) {
+    this.apiUrl = envService.apiRootUrl;
+
     this.loadData();
   }
 
   private loadData() {
-    const employeeUrl =
-      'https://nitro-km-payroll-processor.azurewebsites.net/api/EmployeesGetTrigger';
-
     this.http
-      .get<Employee[]>(employeeUrl)
+      .get<Employee[]>(`${this.apiUrl}/EmployeesGetTrigger`)
       .pipe(
         catchError(err => {
           console.log('Could not fetch employees');
@@ -30,11 +32,8 @@ export class DataService {
       )
       .subscribe(result => (this.employees = result));
 
-    const payrollUrl =
-      'https://nitro-km-payroll-processor.azurewebsites.net/api/PayrollGetTrigger';
-
     this.http
-      .get<Payroll[]>(payrollUrl)
+      .get<Payroll[]>(`${this.apiUrl}/PayrollGetTrigger`)
       .pipe(
         catchError(err => {
           console.log('Could not fetch payroll');
@@ -53,11 +52,17 @@ export class DataService {
   }
 
   removeEmployee(id: string) {
-    this.employees.find(e => e.id === id).status = 'DISABLED';
+    const employee = this.employees.find(e => e.id === id);
+
+    if (!employee) {
+      return;
+    }
+
+    employee.status = 'DISABLED';
+
     this.payroll
       .filter(p => p.employeeId === id)
-      .forEach(payroll => {
-        this.payroll.find(p => p.id === payroll.id).employeeStatus = 'DISABLED';
-      });
+      .filter(payroll => !!this.payroll.find(p => p.id === payroll.id))
+      .forEach(payroll => payroll.employeeStatus = 'DISABLED');
   }
 }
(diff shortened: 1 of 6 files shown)


payroll-processor-client/tsconfig.json
```diff
@@ -20,7 +25,7 @@
     ]
   },
   "angularCompilerOptions": {
-    "fullTemplateTypeCheck": true,
+    "strictTemplates": true,
```
seangwright, line 28: This is the new Angular 9 replacement for `fullTemplateTypeCheck`

https://angular.io/guide/template-typecheck#strict-mode

payroll-processor-client/src/environments/environment.ts
```diff
@@ -1,16 +1,8 @@
-// This file can be replaced during build by using the `fileReplacements` array.
-// `ng build --prod` replaces `environment.ts` with `environment.prod.ts`.
-// The list of file replacements can be found in `angular.json`.
+import { Environment } from './environment-types';
 
-export const environment = {
-  production: false
+export const environment: Environment = {
```
seangwright, line 3: Each `environment` object exported by each `environment.*.ts` file, implements the `Environment` interface to make sure we don't have inconsistent settings between files.

payroll-processor-client/src/app/shared/env.service.ts
```diff
@@ -0,0 +1,16 @@
+import { Injectable } from '@angular/core';
+import { Environment } from '../../environments/environment-types';
+import { environment } from '../../environments/environment';
+
+@Injectable({
+  providedIn: 'root'
+})
+export class EnvService {
```
seangwright, line 8: This abstraction tends to stay pretty small in most apps, but we can use it for some utility access, like the `apiRootUrl` getter below.

payroll-processor-client/src/app/data/data-service.ts
```diff
@@ -53,11 +52,17 @@ export class DataService {
   }
 
   removeEmployee(id: string) {
-    this.employees.find(e => e.id === id).status = 'DISABLED';
```
seangwright, line 56: `find()` can return `undefined`, so I updated this to guard against attempting to access `.status` on `undefined`.

Same with the additional `.filter()` call below.

payroll-processor-client/src/app/data/data-service.ts
```diff
@@ -12,16 +13,17 @@ export class DataService {
   private employees: Employee[] = [];
   private payroll: Payroll[] = [];
 
-  constructor(private http: HttpClient) {
+  private readonly apiUrl: string;
+
+  constructor(private http: HttpClient, envService: EnvService) {
```
seangwright, line 18: Sometimes we won't need the entire service as a dependency, and instead we can grab just the values we need in the constructor.

That's the case here, so I'm just getting the `apiRootUrl` for our api calls.

payroll-processor-client/src/app/data/data-service.ts
```diff
@@ -12,16 +13,17 @@ export class DataService {
   private employees: Employee[] = [];
   private payroll: Payroll[] = [];
 
-  constructor(private http: HttpClient) {
+  private readonly apiUrl: string;
```
seangwright, line 16: For any class fields that won't be changed, I mark them `readonly`.

This helps, as classes grow in complexity, to show exactly which state can change. By limiting what state can change we limit that growth in complexity.

payroll-processor-client/src/environments/environment.ts
KyleMcMaster: Nice! Seeing it with the examples above is a good demonstration.

payroll-processor-client/src/app/data/data-service.ts
KyleMcMaster: I like this, currently a standard practice in my C# code but didn't know it was a feature for properties in TypeScript

payroll-processor-client/src/app/data/data-service.ts
KyleMcMaster: Good catch, this is the type of stuff I am still learning about this language.

payroll-processor-client/src/app/data/data-service.ts
```diff
@@ -53,11 +52,17 @@ export class DataService {
...
+    }
+
+    employee.status = 'DISABLED';
+
     this.payroll
       .filter(p => p.employeeId === id)
-      .forEach(payroll => {
-        this.payroll.find(p => p.id === payroll.id).employeeStatus = 'DISABLED';
-      });
+      .filter(payroll => !!this.payroll.find(p => p.id === payroll.id))
```
KyleMcMaster, line 65: What does this !! do in this expression?

KyleMcMaster:
@all-contributors please add @KyleMcMaster for design, code, and tests

KyleMcMaster:
@all-contributors please add @seangwright  for design and code

KyleMcMaster:
@all-contributors please add @seangwright for design, review, and code

payroll-processor-client/src/app/data/data-service.ts
seangwright: Turning on the strict type checking will help too because the editor and compiler will error in situations like this.

payroll-processor-client/src/app/data/data-service.ts
seangwright: We can also do this with the injected services.

```typescript
private readonly http: HttpClient
```

PR #14, opened 2020-04-24
Feature: table storage / queue integration

seangwright:
This adds data persistence in the Azure functions project using Table Storage / Queue Storage.

It currently relies on the [Azure Storage Emulator](https://docs.microsoft.com/en-us/azure/storage/common/storage-use-emulator) and [Azure Storage Explorer](https://azure.microsoft.com/en-us/features/storage-explorer/) for local development.

Below is an example of some of the backend processes taking place to persist data:

Annotation 2020-04-24 203435 (1)

The data is partially denormalized to get around Azure Table Storage's indexing limitations (partition key + row key).

Data persistence is also eventually consistent, utilizing Queues to queue messages to update related tables to the one directly being modified by a Function request.

Instructions on how to set up a local environment are added to the README.md.

New types were added in the Functions project to support common tasks:

- data access / update
- (de)serialization
- de-normalized data
- data source initialization
- api schema vs db schema


payroll-processor-functions/src/payroll-processor-functions/Features/Payrolls/PayrollData.cs
```diff
@@ -1,7 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 
-namespace PayrollProcessor.Functions.Features.Payroll
+namespace PayrollProcessor.Functions.Features.Payrolls
```
seangwright, line 4: I changed the namespaces to plural to prevent conflicts between class names and namespace names

See: https://stackoverflow.com/a/18731452/939634

payroll-processor-functions/src/payroll-processor-functions/Features/Payrolls/PayrollEntity.cs
```diff
@@ -0,0 +1,81 @@
...
+{
+    public class PayrollEntity : TableEntity
+    {
+        public DateTimeOffset CheckDate { get; set; }
+        public Guid EmployeeId { get; set; }
+        public double GrossPayroll { get; set; }
+        public string PayrollPeriod { get; set; } = "";
+        public string EmployeeDepartment { get; set; } = "";
+
+        public static class Map
```
seangwright, line 14: This is a convenient convention for mapping between entity/domain types.

I added it as a class on the entity type to make it easy to access the write mapper-per-type.

payroll-processor-functions/src/payroll-processor-functions/Features/Payrolls/PayrollTrigger.cs
```diff
@@ -0,0 +1,156 @@
...
+using System;
+using System.Threading.Tasks;
+
+namespace PayrollProcessor.Functions.Features.Payrolls
+{
+    public class PayrollTrigger
+    {
+        [FunctionName("Payroll_GetAll")]
+        public async Task<IActionResult> GetPayrolls(
+            [HttpTrigger(AuthorizationLevel.Anonymous, "GET", Route = "payrolls")] HttpRequest req,
```
seangwright, line 19: I'm using the `Route` property of the `HttpTrigger` attribute to give the Functions endpoints more of a "REST-ish" look and feel.

payroll-processor-functions/src/payroll-processor-functions/Features/Payrolls/PayrollTrigger.cs
```diff
@@ -0,0 +1,156 @@
...
+using PayrollProcessor.Functions.Features.Employees;
+using PayrollProcessor.Functions.Infrastructure;
+using System;
+using System.Threading.Tasks;
+
+namespace PayrollProcessor.Functions.Features.Payrolls
+{
+    public class PayrollTrigger
+    {
+        [FunctionName("Payroll_GetAll")]
```
seangwright, line 17: I should probably use the `nameof()` operator here instead of custom naming the functions different than the method names.

payroll-processor-functions/src/payroll-processor-functions/Features/Payrolls/PayrollTrigger.cs
```diff
@@ -0,0 +1,156 @@
...
+
+            var payrolls = await payrollQuerier.GetAllData<Payroll, PayrollEntity>(e => PayrollEntity.Map.To(e));
+
+            return new OkObjectResult(Response.Generate(payrolls));
+        }
+
+        [FunctionName("EmployeePayrolls_GetAll")]
+        public async Task<IActionResult> GetAllPayrollsForEmployee(
+            [HttpTrigger(AuthorizationLevel.Anonymous, "GET", Route = "employees/{employeeId}/payrolls")] HttpRequest req,
+            [Table("employeePayrolls")] CloudTable employeePayrollsTable,
```
seangwright, line 35: Tables/queues can be accessed by `new`'ing up a table/queue service from the SDK, but the auto-injection that Functions gives us is super convenient.

We miss out on having an easily testable abstraction, but I was able to get around this in the current unit test without _too_ much effort.

payroll-processor-functions/src/payroll-processor-functions/Features/Payrolls/PayrollTrigger.cs
```diff
@@ -0,0 +1,156 @@
...
+            payroll.Id = Guid.NewGuid();
+
+            var querier = new TableQuerier(employeeTable);
+
+            var option = await querier.GetEntity<EmployeeEntity, Employee>(
+                payroll.EmployeeDepartment.ToLowerInvariant(),
+                payroll.EmployeeId.ToString("n"),
+                EmployeeEntity.Map.To);
+
+            var employee = option.IfNone(() => throw new Exception($"Could not find employee [{payroll.EmployeeDepartment}] [{payroll.EmployeeId}]"));
```
seangwright, line 72: I found this `option.IfNone()` to be a nice way to exit early on a failure to get a required value that might not be found in the data persistence layer.

payroll-processor-functions/src/payroll-processor-functions/Features/Payrolls/PayrollTrigger.cs
```diff
@@ -0,0 +1,156 @@
...
+
+            var tableResult = await payrollsTable.ExecuteAsync(TableOperation.Insert(PayrollEntity.Map.From(payroll)));
+
+            if (!(tableResult.Result is PayrollEntity payrollEntity))
+            {
+                throw new Exception($"Could not save payroll");
+            }
+
+            await payrollUpdatesQueue.AddMessageAsync(EntityQueueMessageProcessor.ToQueueMessage(payrollEntity));
```
seangwright, line 81: To keep this function focused on doing 1 thing, I add the updated `PayrollEntity` as a message to the queue. This allows other functions to handle updating the rest of the related entities.

payroll-processor-functions/src/payroll-processor-functions/Features/Resources/ResourcesTrigger.cs
```diff
@@ -0,0 +1,30 @@
...
+using Microsoft.Azure.WebJobs;
+using Microsoft.Azure.WebJobs.Extensions.Http;
+using Microsoft.Extensions.Logging;
+
+namespace PayrollProcessor.Functions.Features.Resources
+{
+    public class ResourcesTrigger
+    {
+        [FunctionName("Resources_Create")]
+        public async Task<IActionResult> CreateEmployeesTable(
```
seangwright, line 14: I like to move this into another project so it can be executed independently of running the functions and so it isn't exposed to the client app.

payroll-processor-functions/src/payroll-processor-functions/host.json
```diff
@@ -6,5 +6,10 @@
         "isEnabled": true
       }
     }
+  },
+  "extensions": {
+    "http": {
+      "routePrefix": "api"
```
seangwright, line 12: I believe this is actually the default for Functions and I could revert this change.

payroll-processor-functions/src/payroll-processor-functions/local.settings.json.sample
```diff
@@ -0,0 +1,11 @@
+{
+  "IsEncrypted": false,
+  "Values": {
+    "AzureTableStorage": "UseDevelopmentStorage=true"
```
seangwright, line 4: This is a convention used for all the injected `[Table]`/`[Queue]` attributes in the functions.

They automatically source their connection from this settings property.

payroll-processor-functions/src/payroll-processor-functions/payroll-processor-functions.csproj
```diff
@@ -4,7 +4,15 @@
     <AzureFunctionsVersion>v3</AzureFunctionsVersion>
     <RootNamespace>Payroll.Processor.Functions</RootNamespace>
   </PropertyGroup>
+
+  <PropertyGroup>
+    <Nullable>enable</Nullable>
```
seangwright, line 9: This turns on C#8 nullable type checking.

payroll-processor.code-workspace
```diff
@@ -52,7 +58,9 @@
       "EditorConfig.EditorConfig",
       "ms-vscode.vscode-typescript-tslint-plugin",
       "dozerg.tsimportsorter",
-      "stylelint.vscode-stylelint"
+      "stylelint.vscode-stylelint",
+      "k--kato.docomment",
```
seangwright, line 62: These are some convenient C#/.NET extensions to help bridge the gap between VS Code and VS

payroll-processor-functions/test/payroll-processor-functions-test/Features/Employees/EmployeesTriggerTests.cs
```diff
@@ -0,0 +1,75 @@
...
+    public static class TableStorageFixtures
+    {
+        public static CloudTable CreateTable(string tableName) =>
+            Substitute.For<CloudTable>(new Uri($"http://127.0.0.1:10002/devstoreaccount1/{tableName}"));
+
+        public static TableQuerySegment<T> CreateSegment<T>(IEnumerable<T> data)
+        {
+            // https://github.com/Azure/azure-storage-net/issues/619#issuecomment-364090291
+
+            var ctor = typeof(TableQuerySegment<T>)
```
seangwright, line 68: This type doesn't have a public constructor, which makes it very hard to create instances of :)

This approach is something another dev came up with and it works well.

payroll-processor-functions/src/payroll-processor-functions/Infrastructure/TableQuerier.cs
```diff
@@ -0,0 +1,66 @@
...
+
+            return segment.Select(mapper);
+        }
+
+        public async Task<IEnumerable<TModel>> GetAllDataByPartitionKey<TModel, TEntity>(
+            Func<TEntity, TModel> mapper,
+            string partitionKey
+        ) where TEntity : ITableEntity, new()
+        {
+            var filter = TableQuery.GenerateFilterCondition(
```
seangwright, line 39: The SDK API uses OData behind the scenes.

I'm still not familiar with how all the querying works or even what the full set of querying options is, but this combination of calls does filter down to a set of entities that all have the same PartitionKey but different RowKeys.

payroll-processor-functions/src/payroll-processor-functions/Infrastructure/EnvironmentSettings.cs
```diff
@@ -0,0 +1,17 @@
+using System;
+using LanguageExt;
+
+namespace PayrollProcessor.Functions.Infrastructure
+{
+    public static class EnvironmentSettings
```
seangwright, line 6: This is a wrapper around configuration access that uses `Option<T>` to force callers to do something if the value isn't found (eg throw an exception)

payroll-processor-functions/src/payroll-processor-functions/Features/Payrolls/PayrollTrigger.cs
```diff
@@ -0,0 +1,156 @@
...
+                payrollEntity.EmployeeDepartment.ToLowerInvariant(),
+                payrollEntity.EmployeeId.ToString("n"),
+                EmployeeEntity.Map.To);
+
+            var employee = option.IfNone(() => throw new Exception($"Could not find employee for payroll [{payroll.Id}]"));
+
+            employee
+                .Payrolls
+                .Find(p => p.Id == payroll.Id)
+                .Match(p =>
```
seangwright, line 135: The way this works is, if the payroll is found in the employee's `Payrolls` set, then update it.

Otherwise, insert it if the length of the employee's `Payrolls` is less than 30. If it's 30 or greater, don't make any changes.

payroll-processor-functions/src/payroll-processor-functions/Features/Payrolls/PayrollEntity.cs
```diff
@@ -0,0 +1,81 @@
...
+                    EmployeeId = payroll.EmployeeId,
+                    GrossPayroll = payroll.GrossPayroll,
+                    PayrollPeriod = payroll.PayrollPeriod,
+                    EmployeeDepartment = payroll.EmployeeDepartment
+                };
+            }
+        }
+    }
+
+    public class EmployeePayrollEntity : TableEntity
```
seangwright, line 44: This entity exists to make it easier to query for Payrolls by Employee, whereas the `PayrollEntity` is used for finding Payrolls by date entered.

It's effectively the same data but separated into 2 tables to make querying possible. This is required because Table storage only allows for 1 indexed set of data (Partition Key + Row Key).

Here the Partition Key is the EmployeeId and the Row Key is the Payroll Id

payroll-processor-functions/src/payroll-processor-functions/Features/Payrolls/PayrollEntity.cs
```diff
@@ -0,0 +1,81 @@
...
+                    PayrollPeriod = entity.PayrollPeriod,
+                    EmployeeDepartment = entity.EmployeeDepartment
+                };
+            }
+
+            public static PayrollEntity From(Payroll payroll)
+            {
+                return new PayrollEntity
+                {
+                    PartitionKey = payroll.CheckDate.ToString("yyyyMMdd"),
```
seangwright, line 32: Here the Partition Key is the date, which should make querying for all Payrolls within a date range possible.

It unfortunately won't make finding all Payrolls of a specific Employee possible (but that's why `EmployeePayrollEntity` exists)

payroll-processor-functions/src/payroll-processor-functions/Features/Payrolls/PayrollEntity.cs
```diff
@@ -0,0 +1,81 @@
...
+using Microsoft.WindowsAzure.Storage.Table;
+
+namespace PayrollProcessor.Functions.Features.Payrolls
+{
+    public class PayrollEntity : TableEntity
+    {
+        public DateTimeOffset CheckDate { get; set; }
+        public Guid EmployeeId { get; set; }
+        public double GrossPayroll { get; set; }
+        public string PayrollPeriod { get; set; } = "";
```
seangwright, line 11: Setting the `= ""` for `string` props to make the nullable reference type checking happy.

payroll-processor-functions/src/payroll-processor-functions/Features/Payrolls/Payroll.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System;
+
+namespace PayrollProcessor.Functions.Features.Payrolls
+{
+    public class Payroll
+    {
+        public Guid Id { get; set; }
```
seangwright, line 7: I should probably make this nullable or create a new `PayrollNew` type that doesn't have the property to make it clearer that the `Id` is generated by the Function, not the client.

payroll-processor-functions/src/payroll-processor-functions/Features/Employees/EmployeeEntity.cs
```diff
@@ -0,0 +1,63 @@
...
+    public class EmployeeEntity : TableEntity
+    {
+        public string Department { get; set; } = "";
+        public DateTimeOffset EmploymentStartedOn { get; set; }
+        public string FirstName { get; set; } = "";
+        public string LastName { get; set; } = "";
+        public string Phone { get; set; } = "";
+        public string Status { get; set; } = "";
+        public string Title { get; set; } = "";
+        public string Payrolls { get; set; } = "";
```
seangwright, line 20: `Payrolls` is a collection of `Payroll` serialized into a string.

The current business logic stores the 30 initially added `Payrolls` to this set, but I should fix that to be the 30 most recent.

payroll-processor-functions/src/payroll-processor-functions/Features/Employees/EmployeeEntity.cs
```diff
@@ -0,0 +1,63 @@
...
+                    Status = entity.Status,
+                    Title = entity.Status,
+                    Payrolls = payrolls,
+                    Version = entity.ETag
+                };
+            }
+
+            public static EmployeeEntity From(Employee employee)
+            {
+                string payrolls = JsonConvert.SerializeObject(employee.Payrolls, DefaultJsonSerializerSettings.JsonSerializerSettings);
```
seangwright, line 44: Here we deserialize the `Payroll` objects out of the string.

payroll-processor-functions/src/payroll-processor-functions/Features/Employees/Employee.cs
```diff
@@ -0,0 +1,32 @@
...
+    {
+        public Guid Id { get; set; }
+        public string Department { get; set; } = "";
+        public DateTimeOffset EmploymentStartedOn { get; set; }
+        public string FirstName { get; set; } = "";
+        public string LastName { get; set; } = "";
+        public string Phone { get; set; } = "";
+        public string Status { get; set; } = "";
+        public string Title { get; set; } = "";
+        public IEnumerable<EmployeePayroll> Payrolls { get; set; } = new EmployeePayroll[] { };
```
seangwright, line 16: Nullable reference type protection

payroll-processor-functions/Payroll Processor.postman_collection.json
```diff
@@ -0,0 +1,156 @@
+{
+	"info": {
+		"_postman_id": "cca9cfc4-7e8e-4bfb-9c39-dcdb76258948",
+		"name": "Payroll Processor",
```
seangwright, line 4: You can load this up in Postman to be able to more easily send requests to the Functions without the client needing to be hooked up, since we don't have form submission in the client yet or default data populating the Azure Tables.

.vscode/settings.json
```diff
@@ -1,7 +0,0 @@
-{
-  "files.exclude": {
-    "payroll-processor-api": true,
```
seangwright, line 3: These were moved to the Workspace configuration

payroll-processor-functions/src/payroll-processor-functions/Features/Employees/EmployeeEntity.cs
```diff
@@ -0,0 +1,63 @@
...
+
+                return new Employee(Guid.Parse(entity.RowKey))
+                {
+                    Department = entity.Department,
+                    EmploymentStartedOn = entity.EmploymentStartedOn,
+                    FirstName = entity.FirstName,
+                    LastName = entity.LastName,
+                    Phone = entity.Phone,
+                    Status = entity.Status,
+                    Title = entity.Status,
```
KyleMcMaster, line 36: Status maps to Title and Status in this method

payroll-processor-functions/src/payroll-processor-functions/Features/Employees/EmployeeEntity.cs
```diff
@@ -0,0 +1,63 @@
...
+                    RowKey = employee.Id.ToString("n"),
+                    Department = employee.Department,
+                    EmploymentStartedOn = employee.EmploymentStartedOn,
+                    FirstName = employee.FirstName,
+                    LastName = employee.LastName,
+                    Phone = employee.Phone,
+                    Status = employee.Status,
+                    Title = employee.Title,
+                    Payrolls = payrolls,
+                    ETag = employee.Version
```
KyleMcMaster, line 58: What does ETag and employee.Version represent?

seangwright: This is a field that comes from Table storage's `TableEntity` class (or the `ITableEntity` interface if you don't want to use inheritance).

It's a 'rowversion' / 'modified date' and Table storage uses a 'last modified wins' versioning scheme (optimistic concurrency).

If we try to save an entity that has an older ETag than the one in Table storage, the update will fail.

payroll-processor-functions/src/payroll-processor-functions/Features/Employees/EmployeeEntity.cs
seangwright: Good catch! Fixed

payroll-processor-functions/src/payroll-processor-functions/local.settings.json.sample
```diff
@@ -0,0 +1,11 @@
...
+  },
+  "Host": {
+    "LocalHttpPort": 7071,
+    "CORS": "*",
+    "CORSCredentials": false
+  }
```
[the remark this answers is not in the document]
seangwright: I copied this from one of the templates that creating a new azure function, with the scaffolding, creates.

I haven't worked with functions much since v1.x, and so much of the config/APIs have changed.

I'm not sure if any of this needs to be explicitly specified.

payroll-processor-functions/src/payroll-processor-functions/Features/Payrolls/PayrollTrigger.cs
seangwright: Updated to use `nameof()`

payroll-processor-functions/src/payroll-processor-functions/Features/Resources/Resource.cs
```diff
@@ -0,0 +1,17 @@
+namespace PayrollProcessor.Functions.Features.Resources
+{
+    public static class Resource
```
seangwright, line 3: This class helps reduced the scattered strings / stringly-typed values throughout the app.

Since the values are mostly used in attributes, we can't get clever with them and use something other than `const string`

seangwright:
After this PR I would like to make a 'rename' PR to change the functions project / folder name from `payroll-processor-functions` to `PayrollProcessor.Functions` because currently, every time I make a new class the generated namespace starts with `payroll-processor-functions.XXXX` and I have to rename it to match the other classes.

payroll-processor-functions/src/payroll-processor-functions/Features/Resources/ResourcesTrigger.cs
KyleMcMaster: Yeah I would agree with this, as functionality increases we may see other more logical groupings of functions that could be broken out.

payroll-processor-functions/src/payroll-processor-functions/host.json
KyleMcMaster: Yeah as far as I know this is the default route. If we need to customize it in the future we can always bring this code back in.

payroll-processor-functions/src/payroll-processor-functions/local.settings.json.sample
KyleMcMaster: So I had to use `"AzureWebJobsStorage": "UseDevelopmentStorage=true"` as an additional key in this configuration.

payroll-processor-functions/src/payroll-processor-functions/payroll-processor-functions.csproj
KyleMcMaster: Can we use the `<LanguageVersion>latest</LanguageVersion>` to bring in all C# 8 features?

payroll-processor-functions/src/payroll-processor-functions/local.settings.json.sample
seangwright: You know what, that's actually what I have in my local settings - I need to sync that back to the sample!

payroll-processor-functions/src/payroll-processor-functions/payroll-processor-functions.csproj
seangwright: I don't think that's enough - we have to opt-in.

https://docs.microsoft.com/en-us/dotnet/csharp/nullable-references#nullable-contexts

Also, I think the property is `<LangVersion>...</LangVersion>`

https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version#edit-the-project-file

payroll-processor-functions/src/payroll-processor-functions/host.json
seangwright: Updated and removed these settings

payroll-processor-functions/src/payroll-processor-functions/Features/Employees/EmployeesTrigger.cs
```diff
@@ -0,0 +1,52 @@
...
+using PayrollProcessor.Functions.Features.Resources;
+using System;
+using System.Linq;
+
+namespace PayrollProcessor.Functions.Features.Employees
+{
+    public class EmployeesTrigger
+    {
+        [FunctionName(nameof(GetEmployees))]
+        public async Task<ActionResult<Employee[]>> GetEmployees(
```
seangwright, line 18: The way `ActionResult<T>` is defined, we can't return an `ActionResult<IEnumerable<T>>` here, so anywhere we'd be returning interfaces, they need to be converted to concrete types.

In the collection of `IEnumerable<T>`, I've turned them all into arrays.

payroll-processor-functions/src/payroll-processor-functions/Startup.cs
```diff
@@ -0,0 +1,17 @@
+using Microsoft.Azure.Functions.Extensions.DependencyInjection;
+using Newtonsoft.Json;
+using PayrollProcessor.Functions;
+using PayrollProcessor.Functions.Infrastructure;
+
+[assembly: FunctionsStartup(typeof(Startup))]
+
+namespace PayrollProcessor.Functions
+{
+    public class Startup : FunctionsStartup
```
seangwright, line 10: We don't currently have much need for function DI, but we can use this class to set that up when we do.

payroll-processor-functions/src/payroll-processor-functions/payroll-processor-functions.csproj
```diff
@@ -4,7 +4,16 @@
...
+    <WarningsAsErrors>CS8600;CS8602;CS8603</WarningsAsErrors>
+  </PropertyGroup>
+
   <ItemGroup>
+    <PackageReference Include="LanguageExt.Core" Version="3.4.13" />
+    <PackageReference Include="Microsoft.Azure.Functions.Extensions" Version="1.0.0" />
```
seangwright, line 15: This package is needed for the Functions "DI" features.

payroll-processor-functions/test/payroll-processor-functions-test/Features/Employees/EmployeesTriggerTests.cs
```diff
@@ -0,0 +1,69 @@
...
+
+            var segment = TableStorageFixtures.CreateSegment(new[] { entity });
+
+            Table
+                .ExecuteQuerySegmentedAsync(Arg.Any<TableQuery<EmployeeEntity>>(), Arg.Any<TableContinuationToken>())
+                .Returns(segment);
+
+            var result = await Trigger.GetEmployees(DefaultHttpRequest, Table, Logger);
+
+            var employeesResult = result.Value;
```
seangwright, line 46: Using `ActionResult<T>` means a much simpler testing process.

payroll-processor-functions/src/payroll-processor-functions/payroll-processor-functions.csproj
KyleMcMaster: Yeah it's LangVersion, was recalling from memory. Good to know!

payroll-processor-functions/src/payroll-processor-functions/local.settings.json.sample
KyleMcMaster: I think this still needs sync'd

payroll-processor-functions/test/payroll-processor-functions-test/Features/Employees/EmployeesTriggerTests.cs
KyleMcMaster: Very creative use of reflection!

payroll-processor-functions/src/payroll-processor-functions/Features/Payrolls/PayrollTrigger.cs
KyleMcMaster: much cleaner!

payroll-processor-functions/src/payroll-processor-functions/Features/Payrolls/PayrollEntity.cs
KyleMcMaster: At some point it would be useful to make a db schema diagram, with the tables and their partition keys for reference.

payroll-processor-functions/src/payroll-processor-functions/Features/Payrolls/PayrollEntity.cs
KyleMcMaster: I like this over using AutoMapper for our domain, since we should only have explicit conversions between types

payroll-processor-functions/src/payroll-processor-functions/Features/Payrolls/PayrollData.cs
KyleMcMaster: Yeah they should have been from the start, i run into this issue all the time with singular namespaces

payroll-processor-functions/src/payroll-processor-functions/Features/Payrolls/Payroll.cs
KyleMcMaster: We can come back to this when the client starts creating payrolls

payroll-processor-functions/src/payroll-processor-functions/Features/Employees/EmployeeEntity.cs
KyleMcMaster: Most recent makes sense to me

payroll-processor-functions/src/payroll-processor-functions/Features/Employees/Employee.cs
KyleMcMaster: Always good to initialize objects, noting you could use `Enumerable.Empty<EmployeePayrolls>();`

payroll-processor-functions/src/payroll-processor-functions/local.settings.json.sample
```diff
@@ -0,0 +1,11 @@
+{
+  "IsEncrypted": false,
+  "Values": {
+    "AzureWebJobsStorage": "UseDevelopmentStorage=true"
```
seangwright, line 4: This should be updated. Maybe the GH ui isn't showing it?

payroll-processor-functions/src/payroll-processor-functions/local.settings.json.sample
seangwright: This was updated but GH isn't updating this comment thread with the latest code... see below.

PR #15, opened 2020-04-25
Add local environment configuration

KyleMcMaster:
Add local configuration environment for client for debugging


README.md
```diff
@@ -87,6 +87,14 @@ and select `Payroll.Processor.Api.sln`.
 
 ### Client
 
+- Copy
```
seangwright, line 90: 👍

seangwright:
Im wondering what's causing the huge file diffs...

.editorconfig
```diff
@@ -2,7 +2,6 @@ root = true
 
 [*]
 charset = utf-8
-end_of_line = crlf
```
seangwright, line 5: I think this was the line-ending issue culprit, combined with your git on Linux not being set to convert `\r\n` to `\n` on commit.

So I removed it and added a `.gitattributes`. I think this should resolve the issue going forward.

payroll-processor-client/package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.0.0",
   "scripts": {
     "ng": "ng",
-    "start": "npm run ng serve -- --aot --port 4201",
+    "start": "npm run ng serve -- --aot --port 4201 --configuration local",
```
seangwright, line 6: There's also a shorthand for this flag - `-c local`

payroll-processor-client/import-sorter.json
```diff
@@ -0,0 +1,30 @@
+{
+  "groupRules": [
```
seangwright, line 2: I had forgotten to add this to the other configuration PRs.

This file works with the `dozerg.tsimportsorter` extension to define how imports get organized and sorted.

It's currently set up to create 5 groups of imports.

1. Angular/RxJs imports
2. 3rd part library imports
3. TSConfig path alias imports (when importing modules from other parts of the app)
4. Relative parent path modules
5. Relative sibling path modules

Whenever we add a new 3rd party lib, we can add it to the `^@datorama|@fortawesome|@ng-bootstrap` list to ensure it gets sorted correctly.

payroll-processor-client/src/app/shared/env.service.ts
```diff
@@ -1,16 +1,17 @@
...
+  providedIn: 'root',
 })
 export class EnvService {
   get env(): Environment {
     return environment;
   }
 
   get apiRootUrl(): string {
-    return `${this.env}/api`;
+    return `${this.env.domain}/api`;
```
seangwright, line 15: Nice catch

payroll-processor-client/src/environments/environment.local.sample.ts
```diff
@@ -0,0 +1,6 @@
+import { Environment } from './environment-types';
+
+export const environment: Environment = {
```
seangwright, line 3: We could "merge" defaults with local settings if we wanted, by rewriting this file as:

```
import { Environment } from './environment-types';
import { environment as envSource } from './environment';

export const environment: Environment = {
  ...envSource,
  production: false,
  domain: 'http://localhost:7071',
};
```

That way we could put the defaults in `environment.ts` and anything not explicitly overwritten in `environment.local.ts` would flow through from there.

I've never done this on a project before, but I have wondered if it would be beneficial...

If we did that, there might be better names for the files, like:

`environment.local.user.sample.ts` - user specific local sample (committed)
`environment.local.user.ts` - user specific local configuration copied from `environment.local.user.sample.ts) (gitignored)
`environmemt.local.ts` - default local (committed)

Thoughts?

payroll-processor-functions/.vscode/tasks.json
```diff
@@ -81,7 +81,7 @@
       "detail": "Runs the Payroll Processor function via Azure Function Core Tools on http://localhost:7071",
       "dependsOn": "Function: Build (Debug)",
       "options": {
-        "cwd": "${workspaceFolder}/src/payroll-processor-functions/bin/Debug/netcoreapp3.1/bin"
+        "cwd": "${workspaceFolder}/src/payroll-processor-functions/bin/Debug/netcoreapp3.1"
```
seangwright, line 84: 👍

.editorconfig
KyleMcMaster: It was a fresh install of git on a new linux machine so i had not yet configured the git environment. This is a good solution moving forward

payroll-processor-client/import-sorter.json
KyleMcMaster: I never knew we could customize imports at the project level. Nice!

payroll-processor-client/src/app/shared/env.service.ts
KyleMcMaster: yeah this took me a minute to figure out why my local url was `localhost:7071/[Object object]/api/....`. Nice to have a local environment set up though.

payroll-processor-client/src/environments/environment.local.sample.ts
KyleMcMaster: I like going the suggested route, this provides a working solution out of the box for the repository but allows users to customize their environment fairly easily.

payroll-processor-functions/.vscode/tasks.json
KyleMcMaster: i think this fix is in all the open PRs lol

payroll-processor-client/package.json
KyleMcMaster: updated!

PR #19, opened 2020-04-26
Table storage front end

KyleMcMaster:
Client implementation of crud around employee entities


payroll-processor-client/src/app/data/data-service.ts
```diff
@@ -25,46 +23,32 @@ export class DataService {
...
-    return this.employees;
-  }
-
   getPayroll(): Payroll[] {
     return this.payroll;
   }
 
-  removeEmployee(id: string) {
-    const employee = this.employees.find((e) => e.id === id);
+  // removeEmployee(id: string) {
```
KyleMcMaster, line 40: TODO: remove commended code once employee status update is developed

payroll-processor-client/src/app/app.module.ts
```diff
@@ -1,37 +1,42 @@
...
 import { EmployeeComponent } from './employee/employee.component';
-import { PayrollComponent } from './payroll/payroll.component';
-import { FontAwesomeModule } from '@fortawesome/angular-fontawesome';
-import { AboutComponent } from './about/about.component';
 import { PayrollListComponent } from './payroll/payroll-list/payroll-list.component';
+import { PayrollComponent } from './payroll/payroll.component';
+import { RiskListComponent } from './risk-list/risk-list.component';
 
 @NgModule({
   declarations: [
```
seangwright, line 21: Two things we can start to look at doing, as this App Module grows:

1. SCAM ([Single Component Angular Modules](https://medium.com/marmicode/your-angular-module-is-a-scam-b4136ca3917b)) - this is a nice way to isolate and clarify what depends on what in an Angular app as it grows.

We'll end up with an Angular module definition for each component, and anywhere we want to re-use that component (eg another component's template), we will import that module. This will bring along all of its dependencies and keep the module definitions readable.

2. [Lazy modules for app features](https://angular.io/guide/lazy-loading-ngmodules).

In the routing definition, we'll use the dynamic import syntax (`import('...').then(...)`) to help Angular split code up into multiple ES Modules.

This will lower the initial payload cost of the app and help define domain boundaries in the app.

payroll-processor-client/src/app/employee/employee-create/employee-create.component.ts
```diff
@@ -0,0 +1,41 @@
...
+    phone: new FormControl(''),
+    title: new FormControl(''),
+  });
+  constructor(private employeeCreateService: EmployeeCreateService) {}
+
+  ngOnInit(): void {}
+
+  create() {
+    const employee: Employee = {
+      id: '00000000-0000-0000-0000-000000000000',
```
seangwright, line 28: Should be able to remove this now

payroll-processor-client/src/app/employee/employee-create/employee-create.service.ts
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  createEmployee(employee: Employee) {
+    console.log(employee);
+    return this.http
+      .post<Employee>(`${this.apiUrl}/employees`, employee)
+      .pipe(
+        tap((detail) => {
+          console.log(detail);
+          this.employeeListStore.upsert(detail.id, {
```
seangwright, line 31: I believe we can pass the employee directly in as the 2nd parameter here:

`this.employeeListStore.upsert(detail.id, detail)`

However, the API doesn't currently return the created employee, instead the list will need to re-query for it.

How do you want to handle that? Should the API return the created entity (it can!) or should we keep everything asynchronous and require the app poll for updates?

(This would be a great place to add SignalR to push an update to the app)

payroll-processor-client/src/app/employee/employee-list/employee-list.component.ts
```diff
@@ -11,20 +15,23 @@ import { faSmileBeam, faSkull } from '@fortawesome/free-solid-svg-icons';
 export class EmployeeListComponent implements OnInit, OnDestroy {
   faSkull = faSkull;
   faSmileBeam = faSmileBeam;
+  employees = of<Employee[]>();
```
seangwright, line 18: If the query is assigned to the class field in the constructor (which I recommend) instead of `ngOnInit()`, then we don't need to pre-assign a value to the `employees` field, and we can mark it as `readonly`.

payroll-processor-client/src/app/employee/employee-list/state/employee-list.model.ts
```diff
@@ -0,0 +1,39 @@
+export interface Employee {
```
seangwright, line 1: We can model the different types in the client like we have in the functions now - `Employee` and `EmployeeNew`.

`EmployeeNew` wouldn't have an `id` or `version` field.

We'll want to add `version` to the `Employee` type.

payroll-processor-client/src/app/employee/employee-list/state/employee-list.service.ts
```diff
@@ -0,0 +1,36 @@
...
+  constructor(
+    private http: HttpClient,
+    envService: EnvService,
+    private employeeListStore: EmployeeListStore,
+  ) {
+    this.apiUrl = envService.apiRootUrl;
+  }
+
+  getEmployees(): Observable<void | Employee[]> {
+    return this.http.get<Employee[]>(`${this.apiUrl}/Employees`).pipe(
```
seangwright, line 25: I think you'll want to call `this.store.setLoading(true)` before calling the API.

Also, we can subscribe here in the service instead of returning an observable to the component.

There's a lot of nuance about subscriptions, but for this scenario where we are making a simple HTTP request, calling `.subscribe()` in the service and pushing the results into the store is a perfectly fine pattern.

Since we'll be using `.subscribe()` here, we can do all our work in the observer we pass to `.subscribe()`

```typescript
.pipe(
    catchError((err) => {
      this.employeeListStore.setError({ message: 'Could not load employees' });
      return of([]);
    })
)
.subscribe({
  next: (employees) => this.employeesListStore.set(employees),
  complete: () => this.employeesListStore.setLoading(false)
});
```

payroll-processor-client/src/app/employee/employee-list/state/employee-list.service.ts
```diff
@@ -0,0 +1,36 @@
...
+
+@Injectable({ providedIn: 'root' })
+export class EmployeeListService {
+  private readonly apiUrl: string;
+  private employees: Employee[];
+
+  constructor(
+    private http: HttpClient,
+    envService: EnvService,
+    private employeeListStore: EmployeeListStore,
```
seangwright, line 19: A lot of times I'll keep the dependency names simple until I need to be more specific.

So, here I'd name this `store` instead of `employeeListStore`.

payroll-processor-functions/src/payroll-processor-functions/payroll-processor-functions.csproj
```diff
@@ -21,7 +21,7 @@
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
     <None Update="local.settings.json">
-      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
```
seangwright, line 24: Was this causing issues?

payroll-processor-client/src/app/employee/employee-list/state/employee-list.model.ts
```diff
@@ -0,0 +1,39 @@
...
+  email: string;
+  employmentStartedOn: string;
+  firstName: string;
+  lastName: string;
+  phone: string;
+  status: Status;
+  title: string;
+}
+
+export type Status = 'ACTIVE' | 'DISABLED';
```
seangwright, line 13: 👍👍

payroll-processor-client/src/app/employee/employee-list/state/employee-list.model.ts
```diff
@@ -0,0 +1,39 @@
+export interface Employee {
+  id: string;
```
seangwright, line 2: We could make some of these `readonly`, like `id` and `version`.

payroll-processor-client/src/app/employee/employee-list/employee-list.component.ts
```diff
@@ -11,20 +15,23 @@ import { faSmileBeam, faSkull } from '@fortawesome/free-solid-svg-icons';
...
 
-  constructor(private dataService: DataService) {}
+  constructor(
+    private employeeListService: EmployeeListService,
```
seangwright, line 21: We can mark these as `readonly` as well:

`private readonly employeeListService: EmployeeListService`.

We can also simplify the names since there are only 2 types being injected:

```typescript
private readonly service: EmployeeListService,
private readonly query: EmployeListQuery
```

payroll-processor-client/src/app/employee/employee-list/employee-list.component.html
```diff
@@ -14,7 +14,7 @@
       [ngClass]="{
         'table-secondary': employee.status === 'DISABLED'
       }"
-      *ngFor="let employee of getEmployees()"
+      *ngFor="let employee of employees | async"
```
seangwright, line 17: 👍

payroll-processor-client/src/app/employee/employee-create/employee-create.service.ts
KyleMcMaster: So after discusson, the flow for now will be to return the detail in a restful manner as we would expect from an api. In the future, the functions will queue up pending employees with a null response causing the client to insert the employee into employee-list store locally and as they are created in functionland, signalR will push updates back to the with the "real" persisted data to display to the client.

payroll-processor-functions/src/payroll-processor-functions/payroll-processor-functions.csproj
KyleMcMaster: nope i didn't mean to check this in, it was a test on the app settings

payroll-processor-client/src/app/employee/employee-list/employee-list.component.ts
KyleMcMaster: fixed!

payroll-processor-client/src/app/employee/employee-list/state/employee-list.model.ts
KyleMcMaster: i agree, employmentStartedOn as well

KyleMcMaster: Sure firstName and email too, since i'm thinking in a relational way email would be a primary key of some sort. And in most cases, last name may change more likely than first name

payroll-processor-client/src/app/employee/employee-list/state/employee-list.service.ts
KyleMcMaster: I wondered about this, I went specific because atleast I was being descriptive, but it felt heavy since in a lot of cases there is only 1 service or 1 store being injected right now

payroll-processor-client/src/app/employee/employee-list/employee-list.component.ts
KyleMcMaster: fixed!

payroll-processor-client/src/app/employee/employee-list/state/employee-list.service.ts
seangwright: In C# I still tend to be specific (probably just habit), but I've started moving away from that in Angular code. It's noisy without much benefit.

I'm more ok with changing the names of class/variables/files as the app grows as opposed to following a specific set of rules from the start.

payroll-processor-client/src/app/employee/employee-create/employee-create.service.ts
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  createEmployee(employee: EmployeeCreate) {
+    // console.log(employee);
+    return this.http
+      .post<EmployeeCreateResponse>(`${this.apiUrl}/employees`, employee)
+      .pipe(
+        tap((detail) => {
+          this.employeeListStore.upsert(detail.rowKey, {
+            id: detail.rowKey,
```
seangwright, line 31: In this case `detail` will be `Employee`, not `EmployeeEntity`, so it will have an `id` and not a `rowKey`

payroll-processor-client/src/app/employee/employee-list/employee-list.component.ts
```diff
@@ -1,30 +1,38 @@
...
+  readonly faSkull = faSkull;
+  readonly faSmileBeam = faSmileBeam;
+  readonly employees: Observable<Employee[]>;
 
-  constructor(private dataService: DataService) {}
+  constructor(
+    private service: EmployeeListService,
+    private query: EmployeeListQuery,
+  ) {
+    this.employees = this.query.selectAll();
```
seangwright, line 24: If you want to get super cool - you can assign the query directly to the field!

```typescript
  readonly faSkull = faSkull;
  readonly faSmileBeam = faSmileBeam;
  readonly employees = this.query.selectAll();
```

payroll-processor-client/src/app/employee/employee-list/employee-list.component.ts
```diff
@@ -1,30 +1,38 @@
...
+  }
 
-  ngOnInit() {}
+  ngOnInit() {
+    this.service.getEmployees().subscribe();
```
seangwright, line 28: You can do this in the constructor instead of `ngOnInit()`.

The best way of thinking about the 2 is `constructor` is for everything to prep the component independent of DOM interactions.

`ngOnInit()` is a lifecycle hook specifically for 'Angular' things, like DOM interactions and manipulation.

Since `this.service.getEmployees().subscribe()` doesn't touch the DOM or other 'Angular' things, you are fine to put it in the constructor.

payroll-processor-client/src/app/data/data-service.ts
```diff
@@ -25,46 +23,32 @@ export class DataService {
 
   private loadData() {
     this.http
-      .get<Employee[]>(`${this.apiUrl}/EmployeesGetTrigger`)
+      .get<Payroll[]>(`${this.apiUrl}/Payrolls`)
```
seangwright, line 26: nit: all the urls in the functions endpoints are `dash-case`

payroll-processor-client/src/app/employee/employee-list/state/employee-list.service.ts
KyleMcMaster: Fixed! This is good insight, although i suspect there is still a lot to learn as far as where to subscribe for http requests.

payroll-processor-client/src/app/employee/employee-list/state/employee-list.model.ts
KyleMcMaster: done!

payroll-processor-client/src/app/employee/employee-list/employee-list.component.ts
seangwright: Eventually what we'll end up doing is creating projections of state inside the query class that are combinations of other projections of state:

```
// query.ts

readonly activeEmployeeNames = this.selectActive(({ firstName, lastName }) => `${firstName} ${lastName}`);

/// component.ts

readonly activeEmployeeNames = this.query.activeEmployeeNames;
```

This lets us encapsulate mapping/projection/aggregation logic inside the Query class, and remove a lot of boilerplate in the component class - no explicit field type, no assignment in constructor, no future mutations of the field.

payroll-processor-client/src/app/employee/employee-list/employee-list.component.ts
KyleMcMaster: image

KyleMcMaster: This is what I feel like you are really telling me with these comments 😆 

image

payroll-processor-client/src/app/employee/employee-list/state/employee-list.service.ts
```diff
@@ -0,0 +1,40 @@
...
+        catchError((err) => {
+          this.store.setError({
+            message: 'Could not load employees',
+          });
+          return of([]);
+        }),
+      )
+      .subscribe({
+        next: (employees) => this.store.set(employees),
+        complete: () => this.store.setLoading(false),
```
seangwright, line 37: The reason for putting the `store.setLoading(false)` here is that is covers both scenarios we could encounter - Success and Failure.

If we didn't put it here, we'd have to either put it in a `tap()` _after_ the `catchError()` or put it in both `catchError()` and a `tap()` before `catchError()`.

By putting our 'clean up' in `complete()`, we know we will always execute that code no matter what happens to the stream of data.

(This only applies to finite/one-time observables, like HTTP requests - for infinite observables, like `query.selectAll()`, we'll do things differently.)

payroll-processor-client/src/app/employee/employee-list/employee-list.component.ts
seangwright: haha no way! The team I'm working with now has been doing Angular for 4-5 years and I leave these comments on their PRs.

There's a lot of nuance to all this stuff. It's not easy.

payroll-processor-client/src/app/employee/employee-list/state/employee-list.service.ts
KyleMcMaster: Let's keep them simple in the angular app and grow more verbose as the app increases in size. I'm okay with different naming conventions in the functions vs client, just keep an eye out on PRs for consistencies sake.

payroll-processor-client/src/app/employee/employee-create/employee-create.service.ts
KyleMcMaster: Yeah this was something i changed to make sure my store update and subscription was working, then as we made design changes i just needed to push up.

payroll-processor-client/src/app/employee/employee-list/employee-list.component.ts
KyleMcMaster: this makes sense since we don't use the query object anywhere else in this class

payroll-processor-client/src/app/employee/employee-list/state/employee-list.service.ts
KyleMcMaster: Again, I didn't know about `next` and `complete` but this makes since with things like the http request pipeline of a web api being able to call `next.Invoke` but not necessarily handling all branches that may execute. Good tip!

payroll-processor-client/src/app/data/data-service.ts
KyleMcMaster: we'll have to keep this in mind if we expand beyond crud

payroll-processor-client/src/app/employee/employee-list/state/employee-list.service.ts
seangwright: Some of this is talked about in [Akita's docs on error handling](https://datorama.github.io/akita/docs/additional/notifications) and on [subscription management](https://netbasal.gitbook.io/akita/other/best-practices#subscription-management).

payroll-processor-client/src/app/employee/employee-list/state/employee-list.service.ts
seangwright: The thing being passed to `.subscribe()` is an [Observer](http://reactivex.io/rxjs/class/es6/MiscJSDoc.js~ObserverDoc.html) and the thing being subscribed to is called an Observable.

Observers are objects with callback functions - they get to "observe" the events coming out of Observables (streams of events) with these callbacks.

Each callback is passed a certain kind of event from the stream and each of the callbacks is optional (we could `.subscribe({ error: err => ... })` and 'Observe' the errors only).

`next` handles each emitted event that isn't an error
`error` handles error events
`complete` handles the closing of the stream of Observables (meaning, no more events are going to be emitted)

payroll-processor-client/src/app/employee/employee-create/employee-create.component.ts
KyleMcMaster: fixed!

seangwright:
LGTM!

PR #20, opened 2020-04-29
Feature: SignalR API integration

seangwright:
### SignalR

Adds SignalR to the API solution and the same VS Code integration for build/debug that the Functions solution has.

When entities are created/updated/deleted in the Functions and those changes need to be propagated out to de-normalized tables, we do this out of band of the client request.

When the updates are all completed we send a message from the function (queue trigger) to the API which pushes a message to the client.

Currently, the design for how these messages look or how the client reacts to them is undetermined.

The API domain projects aren't currently being used. We should consider how we'd want to share code between the Functions and the API.

### EditorConfig for C#

EditorConfig settings for C# were added - we can customize these to our liking.

## Fixes

The Employee entity is now updated correctly to contain the 30 most recent payrolls instead of the first 30 submitted.


api/PayrollProcessor.Api/Properties/launchSettings.json
```diff
@@ -19,12 +19,10 @@
     },
     "PayrollProcessor.Api": {
       "commandName": "Project",
-      "launchBrowser": true,
-      "launchUrl": "risk",
       "environmentVariables": {
         "ASPNETCORE_ENVIRONMENT": "Development"
       },
-      "applicationUrl": "https://localhost:5001;http://localhost:5000"
```
seangwright, line 27: We can add back https bindings for all the projects (and update configuration) - this change was made to get things working.

client/src/app/shared/notification.service.ts
```diff
@@ -0,0 +1,32 @@
...
+      .withUrl(`${this.apiUrl}/hub/notifications`)
+      .build();
+
+    this.hubConnection
+      .start()
+      .then(() => console.log('Connection started'))
+      .catch((err) => console.log('Error while starting connection: ' + err));
+
+    this.hubConnection.on('received', ({ source, message }) => {
+      console.log(source, JSON.parse(message));
```
seangwright, line 29: Not currently doing anything with the SignalR messages - but this would be pretty easy to change.

functions/PayrollProcessor.Functions/Features/Employees/Employee.cs
```diff
@@ -17,9 +19,41 @@ public class Employee
...
+
+                return;
+            }
+
+            /*
+             * Find and (if found) replace the least older of all payrolls
+             * older than the one provided
+             * This keeps the set attached to the Employee as the latest 30 by CheckDate
+             */
+            var leastOlderPayroll = Payrolls
```
seangwright, line 46: I just realized I could do this same thing by adding the `payroll` to the `Payrolls` collection, sorting by date desc and then taking the first 30... 🤔

functions/PayrollProcessor.Functions/PayrollProcessor.Functions.csproj
```diff
@@ -14,6 +14,7 @@
   <ItemGroup>
     <PackageReference Include="LanguageExt.Core" Version="3.4.13" />
     <PackageReference Include="Microsoft.Azure.Functions.Extensions" Version="1.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Http" Version="3.1.3" />
```
seangwright, line 17: For HttpClient DI

functions/PayrollProcessor.Functions/Features/Employees/Employee.cs
KyleMcMaster: Yeah prepending is an interesting solution to this query.

client/src/app/shared/notification.service.ts
KyleMcMaster: logging them for now is fine, i brought in ngx-toastr so we can start toasting when we receive messages

PR #22, opened 2020-04-29
Employee Status Update

KyleMcMaster:
Provide UI functionality to enabled/disable employees


payroll-processor-functions/src/payroll-processor-functions/Features/Employees/EmployeesTrigger.cs
```diff
@@ -48,5 +48,32 @@ public class EmployeesTrigger
...
+            var option = await querier.GetEntity<EmployeeEntity, Employee>(
+                updateParams.Department.ToLowerInvariant(),
+                id.ToString("n"),
+                EmployeeEntity.Map.ToEmployee);
+
+            var employee = option.IfNone(() => throw new Exception($"Could not find employee [{updateParams.Department}] [{id}]"));
+
+            employee.Status = updateParams.Status;
+
+            await employeeTable.ExecuteAsync(TableOperation.InsertOrMerge(EmployeeEntity.Map.From(employee)));
```
seangwright, line 74: I think this returns the updated entity, which will have the updated `ETag` on it.

payroll-processor-client/src/app/employee/employee-list/employee-list.component.ts
```diff
@@ -28,4 +29,25 @@ export class EmployeeListComponent implements OnInit, OnDestroy {
   ngOnInit() {}
 
   ngOnDestroy() {}
+
+  updateEmployeeStatus(id: string) {
+    let employee: Employee;
+
+    this.employees
```
seangwright, line 36: A much easier way to get data out of a store in this situation is to use `this.query.getValue()`.

Usually for rendering / binding state to a component / template, we want to use Observables because they will keep the UI consistent with changes in state.

However, for user generated events (like typed characters, clicks, hovers, ect...), it's ok to use values directly (an in Akita's case, pull them directly out of the store) because we only care about what the app state is at the moment of the user's interaction - not what the state will be like in the future.

payroll-processor-client/src/app/employee/employee-list/state/employee-list.service.ts
```diff
@@ -37,4 +37,20 @@ export class EmployeeListService {
         complete: () => this.store.setLoading(false),
       });
   }
+
+  updateEmployeeStatus(id: string, employee: EmployeeUpdate) {
+    this.store.setLoading(true);
+    return this.http
```
seangwright, line 43: You will generally want to either A) return the `Observable<T>` from the HTTP request or B) `.subscribe()`.

In this case you are returning the `Subscription` (disposable resource) to the caller.

Returning the `Observable<T>` gives the caller more control over how the stream is handled, and the ability to participate in the event stream (react to a new event).

Calling `.subscribe()` in the service means the caller of this method (`updateEmployeeStatus`) will need to be informed about the results of the HTTP request some other way (Akita does this by exposing a Query class which is all events emitted by the store).

payroll-processor-functions/src/payroll-processor-functions/Features/Employees/EmployeesTrigger.cs
KyleMcMaster: fixed!

functions/PayrollProcessor.Functions/Features/Employees/EmployeesTrigger.cs
```diff
@@ -70,7 +70,7 @@ public class EmployeesTrigger
 
             var employee = option.IfNone(() => throw new Exception($"Could not find employee [{updateParams.Department}] [{id}]"));
 
-            employee.Status = updateParams.Status;
+            employee.Status = EmployeeDepartment.Find(updateParams.Status).CodeName;
```
seangwright, line 73: I thought about changing the `Status` property type to be `EmployeeStatus`, but didn't want to add that much strictness just yet. 🤷‍♂️

client/src/app/app.component.scss
```diff
@@ -1,3 +1,9 @@
-nav {
-  background-color: #222;
+// Required for bs4
+@import '~bootstrap/scss/functions';
```
seangwright, line 2: You can add these to `src/styles.scss` instead of the app component scss

client/src/app/employee/employee-list/employee-list.component.html
```diff
@@ -22,20 +22,22 @@
...
+          *ngIf="employee.status === 'Disabled'"
           [icon]="faSkull"
         ></fa-icon>
       </td>
       <td>
         <button
           [ngClass]="'btn btn-outline-warning btn-sm'"
-          [disabled]="employee.status === 'DISABLED'"
-          >Disable</button
+          (click)="updateEmployeeStatus(employee)"
```
seangwright, line 36: I typically name event handlers with an `on` prefix, so `onToggleEmployeeStatus(employee)`.

That when when looking at the methods in the component class you know which ones are being used for event handling in the template and you can organize those together.

client/src/app/employee/employee-list/employee-list.component.ts
```diff
@@ -28,4 +31,22 @@ export class EmployeeListComponent implements OnInit, OnDestroy {
...
+      .updateEmployeeStatus(employee)
+      .pipe(
+        catchError((err) => {
+          console.log(`Could not update employee ${employee.id}`);
+          return throwError(err);
+        }),
+      )
+      .subscribe({
+        next: (detail) => {
+          this.toastr.show('Employee sucessfully updated!');
```
seangwright, line 46: Normally I'd do this kind of stuff in the service and try to keep the component as thin as possible.

Think of the component as a Controller + View in MVC or Controller + JSON in an API.

client/src/app/employee/employee-list/employee-list.component.ts
```diff
@@ -20,6 +21,8 @@ export class EmployeeListComponent implements OnInit, OnDestroy {
   constructor(
     private service: EmployeeListService,
     private query: EmployeeListQuery,
+    private store: EmployeeListStore,
```
seangwright, line 24: The store should only be accessible by the Query and the Service, nothing else.

This is what forces an architecture of 1 way data flow.

Changes go to the -> Service

Service changes the -> Store

Store emits which is exposed by the -> Query

Query passes change/data streams to the -> Component

Component passes a requested state change to the -> Service ... ect

client/src/app/payroll/payroll-list/payroll-list.component.ts
```diff
@@ -1,16 +1,16 @@
...
 
 @Component({
   selector: 'app-payroll-list',
   templateUrl: './payroll-list.component.html',
   styleUrls: ['./payroll-list.component.scss'],
 })
 export class PayrollListComponent {
-  faSkull = faSkull;
-  faSmileBeam = faSmileBeam;
+  readonly faSkull = faSkull;
```
seangwright, line 12: 👍 this is a good pattern that I like to use. `readonly` for everything that doesn't change.

client/src/app/employee/employee-list/employee-list.component.html
KyleMcMaster: fixed

client/src/app/employee/employee-list/employee-list.component.ts
KyleMcMaster: fixed according to the discussion we had this morning, moved all of this code to the service

client/src/app/employee/employee-list/employee-list.component.ts
KyleMcMaster: fixed according to our discussion this morning!

client/src/app/app.component.scss
KyleMcMaster: fixed! good catch. 👍

PR #23, opened 2020-04-30
Refactor: rename project folders

seangwright:
No functional changes - only project / folder renaming.

I think there's only 1 file that Git lost track of, which should help with merging this into any currently open PRs


KyleMcMaster:
LGTM

PR #26, opened 2020-05-05
Feat: Cosmos db 🚀🛰☄🌌

seangwright:
# Description

Azure Table storage was a nice adventure, but its limitations quickly became apparent. Any kind of alternate filtering on a data set required multiple tables with different partition keys, and multi-dimensional filtering (username + created date) was going to be either extremely difficult or impossible.

Azure Table storage is also being migrated out of Azure Storage Accounts and into Cosmos DB - it's future is kinda up in the air.

Since Cosmos DB has a free tier now and a pretty solid local emulator (with cross-platform options), I felt the best way to explore new tech but also give us a standard baseline of functionality to work with, Cosmos would be the best option.

This PR switches the data store from Azure Table Storage to Cosmos Db and does some re-architecture to set us up for being able to add new features to the backend in an obvious and consistent way.

## Solution / Projects

The `/functions` folder has been migrated into `/api` and there is only 1 .NET solution `PayrollProcessor.sln`.

The project naming conventions follow a Clean Architecture pattern
- `PayrollProcessor.Core.Domain` - models and core domain types (commands / queries)
- `PayrollProcessor.Data.Persistence` - Cosmos / Azure Queue integration (command / query handlers)
- `PayrollProcessor.Functions.Api` - Azure functions (out of band data processing, currently running off Queue triggers)
- `PayrollProcessor.Infrastructure.Seeding` - Data seeding
- `PayrollProcessor.Web.Api` - CRUD Web Api for client app interaction and SignalR hubs

All the data persistence code sits behind `IQueryDispatcher` and `ICommandDispatcher` calls.

These calls return `TryAsyncOption<T>` from the [LanguageExt](https://github.com/louthy/language-ext) library to make processing results from these operations more functional and streamlined (this processing pattern can be seen in `PayrollProcessor.Web.Api` Endpoints).

The `PayrollProcess.Functions.Api` project no longer has any CRUD HTTP triggers, besides the ones responsible for populating and resetting the Cosmos and Azure Storage data stores, but it does share the ability to access the Command/Query types in `Core.Domain`.


functions/PayrollProcessor.Functions/Features/Departments/DepartmentsTrigger.cs
```diff
@@ -0,0 +1,129 @@
...
+using System;
+using System.Threading.Tasks;
+using System.Linq;
+using PayrollProcessor.Functions.Features.Employees;
+using PayrollProcessor.Core.Domain.Features.Departments;
+using PayrollProcessor.Functions.Features.Employees.QueueMessages;
+
+namespace PayrollProcessor.Functions.Features.Departments
+{
+    public class DepartmentsTrigger
```
seangwright, line 18: I wonder how we should break up these triggers (or leave them together)? They feel like bulk controllers.

.github/workflows/dotnetcore-api.yml
```diff
@@ -1,5 +1,5 @@
-# build and test PayrollProcessor.Api
-name: dotnet core PayrollProcessor.Api - build
+# build and test PayrollProcessor.Web.Api
+name: dotnet core PayrollProcessor.Web.Api - build
```
seangwright, line 2: Maybe we want to simplify the CI build to just build the solution since all .NET code is contained in it now?

.github/workflows/dotnetcore-functions.yml
```diff
@@ -19,6 +19,6 @@ jobs:
         with:
           dotnet-version: 3.1.101
       - name: Build with dotnet
-        run: dotnet build functions/PayrollProcessor.sln --configuration Release
+        run: dotnet build api/PayrollProcessor.Functions.Api --configuration Release
       - name: Unit Tests
-        run: dotnet test functions/PayrollProcessor.sln
+        run: dotnet test api/PayrollProcessor.Functions.Api.Tests
```
seangwright, line 24: Should we 'test' the whole solution or each application separately?

.github/workflows/npm.yml
```diff
@@ -11,7 +11,7 @@ on:
       - master
 
 env:
-  NODE_VERSION: "10.x" # set this to the node version to use
+  NODE_VERSION: "12.x" # set this to the node version to use
```
seangwright, line 14: Updated to 12.x since there will be things we will start using that aren't supported in 10.x.

api/.vscode/tasks.json
```diff
@@ -1,13 +1,64 @@
 {
   "version": "2.0.0",
   "tasks": [
+    // Solution Tasks
+    {
+      "label": "Solution: Build (Debug)",
```
seangwright, line 6: Now we can either build each application or the whole Solution from tasks.

api/PayrollProcessor.Core.Domain/Features/Departments/DepartmentPayrollCreateCommand.cs
```diff
@@ -0,0 +1,25 @@
...
+using PayrollProcessor.Core.Domain.Features.Employees;
+using PayrollProcessor.Core.Domain.Intrastructure.Operations.Commands;
+
+namespace PayrollProcessor.Core.Domain.Features.Departments
+{
+    public class DepartmentPayrollCreateCommand : ICommand<DepartmentPayroll>
+    {
+        public Employee Employee { get; }
+        public EmployeePayroll EmployeePayroll { get; }
+        public Guid RecordId { get; }
```
seangwright, line 12: Couldn't think up a good name for this, but I wanted the caller of the command to supply the new `Guid` that will be sent to Cosmos, as opposed to it being generated internally.

This could let us move from returning the new entity directly from the call (`ICommand<DepartmentPayroll>`) to only returning 'success' (`ICommand`) and having more of a disconnected architecture with Commands not returning values.

api/PayrollProcessor.Core.Domain/Features/Departments/DepartmentPayrollsQuery.cs
```diff
@@ -0,0 +1,37 @@
...
+        {
+            Guard.Against.NullOrWhiteSpace(department, nameof(department));
+
+            Department = department;
+            CheckDateFrom = checkDateFrom;
+            CheckDateTo = checkDateTo;
+            Count = count;
+        }
+
+        public void Deconstruct(out int count, out string department, out DateTime? checkDateFrom, out DateTime? checkDateTo)
```
seangwright, line 29: These methods are unnecessary, but I kinda like them.

api/PayrollProcessor.Core.Domain/Features/Employees/EmployeeCreateCommand.cs
```diff
@@ -0,0 +1,21 @@
+using System;
+using Ardalis.GuardClauses;
+using PayrollProcessor.Core.Domain.Intrastructure.Operations.Commands;
+
+namespace PayrollProcessor.Core.Domain.Features.Employees
+{
+    public class EmployeeCreateCommand : ICommand<Employee>
+    {
+        public Guid NewId { get; set; }
```
seangwright, line 9: Probably need a better name for this property.

api/PayrollProcessor.Core.Domain/Intrastructure/Clocks/DateTimeProvider.cs
```diff
@@ -0,0 +1,17 @@
+using System;
+
+namespace PayrollProcessor.Core.Domain.Intrastructure.Clocks
+{
+    public interface IDateTimeProvider
```
seangwright, line 5: I didn't feel like creating separate files for these Interface abstractions that are so closely tied to their single implementations.

api/PayrollProcessor.Core.Domain/Intrastructure/Identifiers/EntityIdGenerator.cs
```diff
@@ -0,0 +1,14 @@
+using System;
+
+namespace PayrollProcessor.Core.Domain.Intrastructure.Identifiers
+{
+    public interface IEntityIdGenerator
+    {
+        Guid Generate();
+    }
+
+    public class EntityIdGenerator : IEntityIdGenerator
```
seangwright, line 10: This will let us be able test methods that perform Guid creation a little easier.

api/PayrollProcessor.Core.Domain/Intrastructure/Operations/Commands/ICommand.cs
```diff
@@ -0,0 +1,6 @@
+namespace PayrollProcessor.Core.Domain.Intrastructure.Operations.Commands
+{
+    public interface ICommand { }
```
seangwright, line 3: What's nice about baking the `TryOptionAsync<T>` into the `ICommandHandler` is that I don't need to expose it in the `ICommand`.

api/PayrollProcessor.Core.Domain/Intrastructure/Operations/Commands/ICommandDispatcher.cs
```diff
@@ -0,0 +1,12 @@
+using System.Threading;
+using LanguageExt;
+
+namespace PayrollProcessor.Core.Domain.Intrastructure.Operations.Commands
+{
+    public interface ICommandDispatcher
+    {
+        TryOptionAsync<Unit> Dispatch(ICommand command, CancellationToken token = default);
```
seangwright, line 8: `TryOptionAsync<T>` requires a generic type parameter because in functional programming `void` is not a real return type.

`LanguageExt` provides `Unit` which is a return type that implies no value but side effects did occur.

It would be really interesting to also use this approach on the client side - instead of a `: void` return type on Akita service methods, we could use a `: Unit` return type using something like https://gcanti.github.io/fp-ts/

api/PayrollProcessor.Core.Domain/Intrastructure/Operations/Commands/ICommandDispatcher.cs
```diff
@@ -0,0 +1,12 @@
...
+
+        TryOptionAsync<TResponse> Dispatch<TResponse>(ICommand<TResponse> command, CancellationToken token = default);
```
seangwright, line 10: I fixed the issue with previous attempts at these dispatchers/handlers not passing through the `CancellationToken`.

This time all of them pass it through (though the Ardalis.Endpoints doesn't accept it in the method signature, so we'll need to PR into that library to get cancellation added from end-to-end).

api/PayrollProcessor.Core.Domain/Intrastructure/Serialization/DefaultJsonSerializerSettings.cs
```diff
@@ -1,11 +1,11 @@
 ﻿using Newtonsoft.Json;
 using Newtonsoft.Json.Serialization;
 
-namespace PayrollProcessor.Functions.Infrastructure
+namespace PayrollProcessor.Core.Domain.Infrastructure.Serialization
```
seangwright, line 4: I put this in the Domain to make sure we have consistent serialization configuration throughout the backend.

api/PayrollProcessor.Core.Domain/PayrollProcessor.Core.Domain.csproj
```diff
@@ -0,0 +1,13 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.1</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Ardalis.GuardClauses" Version="1.5.0" />
+    <PackageReference Include="LanguageExt.Core" Version="3.4.14" />
```
seangwright, line 9: I'm not sure how I feel about having the Domain project depend on multiple packages.

One one hand, they are all just types/config and not services or external / volatile dependencies.

On the other hand I wonder if a `Core.Processes` project would be a good idea for the CQRS types, `IDateProvider`, ect...

I don't think we need to do anything about it now, but it's interesting to think about 🤔.

api/PayrollProcessor.Data.Persistence/Features/Departments/DepartmentEmployeeCreateCommandHandler.cs
```diff
@@ -0,0 +1,35 @@
...
+            Guard.Against.Null(client, nameof(client));
+
+            this.client = client;
+        }
+
+        public TryOptionAsync<DepartmentEmployee> Execute(DepartmentEmployeeCreateCommand command, CancellationToken token)
+        {
+            var record = DepartmentEmployeeRecord.Map.CreateNewFrom(command.Employee, command.RecordId);
+
+            return async () =>
```
seangwright, line 25: `TryOptionAsync<T>` is a representation of a Func that returns a `Task` of a Some<T>/None, though that Func could also completely fail when executed and throw an Exception.

So here we return the callback that when executed could end up in 1 of 3 states

- T - our value returned by executing the callback
- None - no exceptions were thrown but also no values were returned
- Exception - executing the callback caused an exception to be thrown

api/PayrollProcessor.Data.Persistence/Features/Departments/DepartmentEmployeeRecord.cs
```diff
@@ -0,0 +1,42 @@
...
+{
+    public class DepartmentEmployeeRecord : CosmosDBRecord
+    {
+        public Guid EmployeeId { get; set; }
+        public string Department { get; set; } = "";
+        public string FirstName { get; set; } = "";
+        public string LastName { get; set; } = "";
+        public string Email { get; set; } = "";
+
+        public static class Map
```
seangwright, line 16: I'm finding these nested `Map` classes to be an interesting alternative to `AutoMapper`.

Our domain/persistence layers are pretty similar right now, so AutoMapper would likely handle the mapping fine without much ceremony.

There's something nice about the explicit `Map` methods/functions compared to some of the magic of `AutoMapper`.

Thoughts?

api/PayrollProcessor.Data.Persistence/Features/Departments/DepartmentPayrollsQueryHandler.cs
```diff
@@ -0,0 +1,67 @@
...
+            Guard.Against.Null(client, nameof(client));
+
+            this.client = client;
+        }
+
+        public TryOptionAsync<IEnumerable<DepartmentPayroll>> Execute(DepartmentPayrollsQuery query, CancellationToken token = default)
+        {
+            var (count, department, checkDateFrom, checkDateTo) = query;
+
+            var dataQuery = client
```
seangwright, line 30: I could put this 'setup' code inside the callback returned by this method. That would allow the `Try` to catch any exceptions thrown before the actually data access operation.

api/PayrollProcessor.Data.Persistence/Features/Employees/EmployeeDetailQueryHandler.cs
```diff
@@ -0,0 +1,84 @@
...
+
+            return async () =>
+            {
+                while (iterator.HasMoreResults)
+                {
+                    var response = await iterator.ReadNextAsync(token);
+
+                    foreach (var item in response)
+                    {
+                        string type = item.Value<string>("type") ?? "";
```
seangwright, line 45: We store multiple schemes/shapes/types in the same collection.

The 'discriminator' between them is the `type` property, which the `*Record` types populate through the `.Map` methods.

here we want to query for all values matching a partition key, regardless of what `type` they are (both Employees and EmployeePayrolls).

I looked for examples of how to do this but couldn't find any.

The solution here that I came up with was to delay converting the parsed JSON into a POCO until we know the value of `type`.

An alternative would be to `item.ToObject<CosmosDbRecord>()` and read its strongly typed `Type` property, but the call `item.Value<string>("type") ?? ""` does the exact same thing without the cost of an extra `ToObject<>()` call for a value that gets thrown away.

api/PayrollProcessor.Data.Persistence/Features/Employees/EmployeesQueryHandler.cs
```diff
@@ -0,0 +1,69 @@
...
+            }
+
+            if (!string.IsNullOrWhiteSpace(lastName))
+            {
+                dataQuery = dataQuery.Where(e => e.LastNameLower.Contains(lastName.ToLowerInvariant()));
+            }
+
+            if (count > 0)
+            {
+                dataQuery = dataQuery.Take(count);
```
seangwright, line 51: I found out through testing that the order of LINQ expressions matters with Cosmos.

The `.Take()` call has to be the last, otherwise the `CosmosClient` interprets the LINQ expressions that follow as being part of a sub-query, which isn't allowed to be combined with `.Take()`.

api/PayrollProcessor.Data.Persistence/Infrastructure/Clients/CosmosClientExtensions.cs
```diff
@@ -0,0 +1,33 @@
+using System.Linq;
+using PayrollProcessor.Data.Persistence.Features.Departments;
+using PayrollProcessor.Data.Persistence.Features.Employees;
+
+namespace Microsoft.Azure.Cosmos
+{
+    public static class CosmosClientExtensions
```
seangwright, line 5: These are some utilities to remove some boilerplate when working with the `CosmosClient`

api/PayrollProcessor.Data.Persistence/Infrastructure/Clients/QueueClientFactory.cs
```diff
@@ -0,0 +1,39 @@
...
+using Newtonsoft.Json;
+
+namespace PayrollProcessor.Data.Persistence.Infrastructure.Clients
+{
+    public interface IMessage
+    {
+        string Source { get; set; }
+    }
+
+    public static class QueueMessageBuilder
```
seangwright, line 15: I'm not completely happy with this way of integrating with sending Queue messages, but I don't think it makes sense to try to come up with something better until we use these more and find problems with them.

api/PayrollProcessor.Functions.Api/Features/Resources/ResourcesTrigger.cs
```diff
@@ -0,0 +1,101 @@
...
+        public async Task<ActionResult> CreateResources(
+            [HttpTrigger(AuthorizationLevel.Anonymous, "POST", Route = "resources")] HttpRequest req,
+            ILogger log)
+        {
+            log.LogInformation($"Creating all tables and queues: [{req}]");
+
+            var dbResponse = await client.CreateDatabaseIfNotExistsAsync(Databases.PayrollProcessor.Name);
+
+            await dbResponse.Database.CreateContainerIfNotExistsAsync(
+                new ContainerProperties(Databases.PayrollProcessor.Containers.Employees, partitionKeyPath: "/partitionKey"));
```
seangwright, line 41: Every collection in Cosmos needs a partition key - this is a property on the JSON documents that Cosmos uses to group data together (for horizontal scale).

Cosmos is able to query on items in the same partition of a collection (identified by the partition key property value) compared to querying across multiple partitions in the same collection.

I saw a recommendation to partition every collection on `"partitionKey"` and then add a property to every Document inserted into Cosmos called `partitionKey` instead of each collection having its own unique partition property (like `customerId`, `departmentName`, `creationDate`).

api/PayrollProcessor.Functions.Api/Infrastructure/ApiClient.cs
```diff
@@ -20,17 +24,22 @@ public ApiClient(HttpClient client)
...
-                Message = JsonConvert.SerializeObject(data, DefaultJsonSerializerSettings.JsonSerializerSettings)
+                Message = JsonConvert.SerializeObject(data)
             };
 
-            var response = await client.PostAsJsonAsync("/api/notification", notification);
+            return async () =>
+            {
+                var response = await client.PostAsJsonAsync("/api/notification", notification);
+
+                response.EnsureSuccessStatusCode();
```
seangwright, line 39: This call with throw and exception if the status code was not a success

api/PayrollProcessor.Functions.Api/Infrastructure/QueueMessageHandler.cs
```diff
@@ -0,0 +1,12 @@
+using Microsoft.WindowsAzure.Storage.Queue;
+using Newtonsoft.Json;
+using PayrollProcessor.Data.Persistence.Infrastructure.Clients;
+
+namespace PayrollProcessor.Functions.Api.Infrastructure
+{
+    public class QueueMessageHandler
+    {
+        public static TMessage FromQueueMessage<TMessage>(CloudQueueMessage message) where TMessage : IMessage =>
```
seangwright, line 9: Azure Functions uses `CloudQueueMessage`, whereas the latest Azure Queue library from Microsoft uses the type `QueueMessage`.

The Azure Functions bindings for Azure Storage are slightly out of date.

This means we can't share the same APIs/abstractions in the `Data.Persistence` library and the `Functions.Api` application.

api/PayrollProcessor.Web.Api/Features/Departments/DepartmentEmployeesGet.cs
```diff
@@ -0,0 +1,60 @@
...
+
+    public class DepartmentEmployeesResponse : IListResponse<DepartmentEmployee>
+    {
+        public IEnumerable<DepartmentEmployee> Data { get; }
+
+        public DepartmentEmployeesResponse(IEnumerable<DepartmentEmployee> data)
+        {
+            Guard.Against.Null(data, nameof(data));
+
+            Data = data;
```
seangwright, line 57: All collection responses will now return the collection on a `data` property.

ex:

```json
{
  "data": [{
    "id": 1
    "name": "test"
  }, {
    "id": 2,
    "name": "other"
  }]
}
```

api/PayrollProcessor.Web.Api/Startup.cs
```diff
@@ -0,0 +1,92 @@
...
+
+            services.AddCors(options =>
+                options.AddPolicy("CorsPolicy",
+                    builder => builder
+                        .WithOrigins(Configuration.GetValue<string>("CORS:client:domain"))
+                        .AllowAnyMethod()
+                        .AllowAnyHeader()
+                        .AllowCredentials()));
+
+            services.AddControllers(options => options.UseGlobalRoutePrefix("api/v{version:apiVersion}"));
```
seangwright, line 37: This allows us to not have to type `api/v1` on every endpoint.

client/import-sorter.json
```diff
@@ -21,10 +21,10 @@
       "level": 19
     }
   ],
-  "maximumLineLength": 120,
-  "maximumBindingNamesPerLine": 8,
-  "maximumDefaultAndBindingNamesPerLine": 8,
-  "maximumNamesPerWrappedLine": 4,
-  "formatOnSave": true,
+  "maxLineLength": 120,
```
seangwright, line 24: This extension recently updated and renamed some of these properties.

api/PayrollProcessor.Data.Persistence/Features/Departments/DepartmentEmployeeRecord.cs
KyleMcMaster: I think this works great as an alternative to automapper, expecially since we know exactly which types we will be mapping to and from. No dynamic mapping for us.

api/PayrollProcessor.Web.Api/Features/Employees/EmployeeCreate.cs
```diff
@@ -0,0 +1,70 @@
...
+        {
+            var command = new EmployeeCreateCommand(
+                generator.Generate(),
+                new EmployeeNew
+                {
+                    Department = request.Department,
+                    Email = request.Email,
+                    EmploymentStartedOn = request.EmploymentStartedOn,
+                    FirstName = request.FirstName,
+                    LastName = request.FirstName,
```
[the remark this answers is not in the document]
KyleMcMaster: fixed

api/PayrollProcessor.Web.Api/Startup.cs
```diff
@@ -0,0 +1,95 @@
...
+
+            services.AddApiVersioning(options =>
+            {
+                options.ReportApiVersions = true;
+                options.DefaultApiVersion = new ApiVersion(1, 0);
+            });
+
+            services.AddSignalR();
+
+            services
```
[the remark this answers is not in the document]
KyleMcMaster: fixed, if we have more services like this it may be useful to add a extension method for registering application services

api/PayrollProcessor.Data.Persistence/Features/Employees/EmployeesQueryHandler.cs
KyleMcMaster: Good to know, i'll have to play around with the queries to gain more understanding of them.

api/PayrollProcessor.Functions.Api/Features/Departments/DepartmentEmployeeTrigger.cs
```diff
@@ -0,0 +1,96 @@
...
+
+        private Task HandleEmployeeCreation(CloudQueueMessage queueMessage, ILogger log)
+        {
+            var message = QueueMessageHandler.FromQueueMessage<EmployeeCreation>(queueMessage);
+
+            return queryDispatcher.Dispatch(new EmployeeQuery(message.EmployeeId))
+                .Bind(employee => commandDispatcher.Dispatch(new DepartmentEmployeeCreateCommand(employee, idGenerator.Generate())))
+                .Bind(departmentEmployee => apiClient.SendNotification(nameof(EmployeeUpdatesQueue), departmentEmployee))
+                .Match(
+                    _ => log.LogInformation(""),
```
seangwright, line 75: Definitely need to add some useful logging here and in the `DepartmentPayrollTrigger` as well.

PR #28, opened 2020-05-11
Payroll list

KyleMcMaster:
Payroll list with side nav for filtering by department.


client/src/app/payroll/payroll-list/payroll-list.component.html
```diff
@@ -4,18 +4,15 @@
...
-    <tr
-      [ngClass]="{
-        'table-secondary': payroll.employeeStatus === 'Disabled'
-      }"
-      *ngFor="let payroll of getPayroll()"
-    >
-      <th scope="row"> {{ payroll.employeeName }} </th>
+    <tr *ngFor="let payroll of payrolls | async">
+      <th scope="row">
+        {{ payroll.employeeFirstName + ' ' + payroll.employeeLastName }}
```
seangwright, line 14: You can also `{{ payroll.employeeFirstName }} {{ payroll.employeeLastName }}`

client/src/app/payroll/payroll-list/state/payroll-list.service.ts
```diff
@@ -0,0 +1,52 @@
...
+        {
+          params,
+        },
+      )
+      .pipe(
+        catchError((err) => {
+          this.store.setError({
+            message: 'Could not load payrolls',
+          });
+          return of([]);
```
seangwright, line 43: `return EMPTY` is a pattern I like more than `of([])`, because `EMPTY` completes immediately.

client/src/app/payroll/payroll-list/state/payroll-list.service.ts
```diff
@@ -0,0 +1,52 @@
...
+        }),
+        mapListResponseToData(),
```
seangwright, line 45: You can move this above `catchError()` when using `EMPTY` instead of `of()`

client/src/app/payroll/payroll.component.html
```diff
@@ -1 +1,39 @@
...
+          </li>
+          <li ngbNavItem="Warehouse">
+            <a ngbNavLink>Warehouse</a>
+          </li>
+        </ul>
+      </span>
+      <div [ngbNavOutlet]="nav" class="ml-4"></div>
+    </div>
+    <div class="col-10">
+      <app-payroll-list department="{{ department }}"></app-payroll-list>
```
seangwright, line 36: This works, though there is a more common syntax for binding values to component `@Input()`s.

```
<app-payroll-list [department]="department"></app-payroll-list>
```

Here `[department]` binds to the component property (JS) instead of the component attribute (HTML).

The binding syntax `[someProp]` can be assigned any JS expression - in this case the value of the parent component `department` field.

client/src/app/payroll/payroll-list/state/payroll-list.model.ts
```diff
@@ -0,0 +1,25 @@
+import { Department } from 'src/app/employee/employee-list/state/employee-list.model';
+
+export interface PayrollList {
+  id: string;
+  checkDate: Date;
```
seangwright, line 5: This is technically a `string` when it comes from the API since there is no `Date` type in JSON.

If you want to have a `Date` type internally, you can separate the representation of API types from APP types - kinda like separating Records (persistence) from Entities (domain).

What I often do is create a `Schema` type for what the API responds with and a different type for what the app works with.

Then I create functions to map between the `Schema` and app types.

When getting a response back from the API using the `HttpClient` I do something like:

```typescript
// mapping function

export function mapPayrollSchemaToPayroll(schema: PayrollSchema): Payroll {
  return {
    // map props ...
    checkDate: new Date(schema.checkDate),

    // map remaining props ...
  }
}

// service class

.pipe(
  mapListResponseToValues(mapPayrollSchemaToPayroll)
)
```

client/src/app/payroll/payroll-list/state/payroll-list.service.ts
```diff
@@ -0,0 +1,52 @@
...
+import { PayrollListStore } from './payroll-list.store';
+
+@Injectable({ providedIn: 'root' })
+export class PayrollListService {
+  readonly apiRootUrl: string;
+
+  constructor(
+    envService: EnvService,
+    private http: HttpClient,
+    protected store: PayrollListStore,
```
seangwright, line 20: You can make all of these `private readonly`

client/src/app/payroll/payroll-list/state/payroll-list.service.ts
```diff
@@ -0,0 +1,52 @@
...
+  ) {
+    this.apiRootUrl = envService.apiRootUrl;
+  }
+
+  getPayrolls(department: string) {
+    this.store.setLoading(true);
+
+    let params = new HttpParams();
```
seangwright, line 28: nit `const`

client/src/app/payroll/payroll-list/state/payroll-list.service.ts
```diff
@@ -0,0 +1,52 @@
...
+      .get<ListResponse<PayrollList>>(
+        `${this.apiRootUrl}/departments/payrolls`,
+        {
+          params,
+        },
+      )
+      .pipe(
+        catchError((err) => {
+          this.store.setError({
+            message: 'Could not load payrolls',
```
seangwright, line 41: 👍

client/src/app/payroll/payroll-list/state/payroll-list.service.ts
KyleMcMaster: Good to know!

client/src/app/payroll/payroll-list/payroll-list.component.html
KyleMcMaster: fixed

client/src/app/payroll/payroll-list/state/payroll-list.model.ts
KyleMcMaster: Right now the client doesn't perform any logic on checkDate so i think changing it to a string for now is fine, if we end up needing to use the date for logic in the app, we can use the schema mapping concept then.

client/src/app/payroll/payroll.component.html
seangwright: One thing you'll often see is when needing to bind a literal string value to something like `@Input name: string`, instead of writing `<a-component [name]="'Sean'"></a-component>`, you can use the HTML attribute and write `<a-component name="Sean"></a-component>`, which lets you avoid the quotes.

HTML attributes require strings (`name="Sean"`) whereas JavaScript properties require expressions (`[name]="'Sean'"`).

client/src/app/app.component.html
```diff
@@ -15,47 +15,16 @@
               <app-employee></app-employee>
             </ng-template>
           </li>
-          <li [ngbNavItem]="2">
-            <a ngbNavLink>Payroll</a>
+          <li [ngbNavItem]="'payrolls'">
```
seangwright, line 18: Here's a good place to use the attribute binding instead of the property binding so you do need to use nested quotes:

`<li ngbNavItem="payrolls">`

KyleMcMaster:
Abandoning this PR but will implement this functionality elsewhere in the app

PR #30, opened 2020-05-29
Refactor employee UI structure

KyleMcMaster:
Use a parent component to get, organize, and push data to employee list and employee detail components.


client/src/app/employee/employee-create/employee-create.component.html
```diff
@@ -1,82 +1,82 @@
...
+        <input
+          type="date"
+          formControlName="employmentStartedOn"
+          class="form-control"
+          placeholder="04/26/2020"
+          aria-label="Users Start Date Of Employment"
+        />
+      </div>
+      <div class="mb-1 col-4">
+        <button class="btn btn-outline-info table-margin" (click)="create()"
```
seangwright, line 76: To align with HTML form semantics, you can move this event handler to the `(submit)` event of the `<form>`. Then change this button to `<button type="submit" ...>`, or leave off the `type="submit"` because that's the default for `<button>`.

client/src/app/employee/employee-detail/employee-detail.component.html
```diff
@@ -0,0 +1,62 @@
+<ng-container *ngIf="employee | async; let employee" class="border-info">
```
seangwright, line 1: The unwrapping (`| async`) and displaying of this data _could_ be combined in the same component because the behavior is pretty simple here, however this is usually the boundary that you want to use to separate off your presentational components, which are responsible for laying out state in markup, but only get that state via `@Input()`, not services or `Observable<>`

client/src/app/employee/employee-detail/employee-detail.component.ts
```diff
@@ -0,0 +1,17 @@
...
+import { Employee } from './state/employee-detail.model';
+
+@Component({
+  selector: 'app-employee-detail',
+  templateUrl: './employee-detail.component.html',
+  styleUrls: ['./employee-detail.component.scss'],
+})
+export class EmployeeDetailComponent implements OnInit {
+  @Input()
+  employee: Observable<Employee>;
```
seangwright, line 12: This should probably be `@Input() employee: Employee` and the `| async` should happen in the parent component passing the state in.

client/src/app/employee/employee-detail/state/employee-detail.store.ts
```diff
@@ -0,0 +1,13 @@
+import { Injectable } from '@angular/core';
+import { EntityState, Store, StoreConfig } from '@datorama/akita';
+import { createInitialState, Employee } from './employee-detail.model';
+
+export interface EmployeeDetailState extends EntityState<Employee> {}
+
+@Injectable({ providedIn: 'root' })
+@StoreConfig({ name: 'employee-detail' })
+export class EmployeeDetailStore extends Store<EmployeeDetailState> {
```
seangwright, line 9: Did you make these with the Angular schematics?

If your state extends `EntityState<T>` then you should be using `EntityStore` and `QueryEntity`.

Then you could create a custom `Observable<>` using the `selectActive()` method that selected the first active employee.

client/src/app/employee/employee-list/employee-list.component.ts
```diff
@@ -13,21 +22,21 @@ import { EmployeeListService } from './state/employee-list.service';
...
 
-  constructor(
-    private service: EmployeeListService,
-    private query: EmployeeListQuery,
-  ) {
-    this.employees = this.query.selectAll();
-    this.service.getEmployees();
-  }
+  @Input()
+  employees: Observable<EmployeeList[]>;
```
seangwright, line 28: You probably want `@Input() employees: EmployeeList[]` here, not an `Observable<>`

client/src/app/employee/employee-list/state/employee-list.model.ts
```diff
@@ -1,32 +1,22 @@
-export interface Employee {
+export interface EmployeeList {
```
seangwright, line 1: This reads, to me, like a custom type alias for `Employee[]`, but instead its the shape of employee data for the `employees` endpoint.

Maybe there's a better name?

In other apps I've worked on we've used `ListEmployee` for this, or `Employee` and `EmployeeDetail` to represent the minimal and full schemas.

client/src/app/employee/employee.component.ts
```diff
@@ -10,11 +16,30 @@ import { faLock, faUnlock } from '@fortawesome/free-solid-svg-icons';
...
+  }
+
+  setCreate() {
+    this.uiState = 'create';
+  }
 
-  toggleReadonly() {
-    this.isReadonly = !this.isReadonly;
+  setActive(employee: EmployeeList) {
+    this.employeeDetailService.getEmployee(employee.id);
```
seangwright, line 42: These two operations would go well together. These are two state changes (setting the `uiState` and populating the `Employee`) that necessarily both have to happen for the application to work correctly.

This is an example where moving local state into an Akita store and having the `EmployeeDetailService` both request the new `Employee` data and change the `uiState` in the store, would be a good approach.

It's fine to leave it as it is for now though!

client/src/app/employee/employee-detail/employee-detail.component.html
KyleMcMaster: So in this example the `EmployeeComponent` would subscribe the the `Observable<Employee>` and push that Employee object into the `EmployeeDetailComponent`?

client/src/app/employee/employee-detail/state/employee-detail.store.ts
KyleMcMaster: I did make them with the Angular Schematics, not 100% sure i made the right decisions.

client/src/app/employee/employee-detail/employee-detail.component.ts
KyleMcMaster: So Justin taught me something new today, I never realized you could async pipe an input like `[employee]="this.employee | async"`. So now this makes more sense on how to have the parent component manage gathering all the data, handle any observables, and then push that data into "dumb" child components.

client/src/app/employee/employee-list/employee-list.component.ts
KyleMcMaster: fixed!

client/src/app/employee/employee.component.ts
KyleMcMaster: Definitely want to come back to this at some point, i haven't had an opportunity to store the UI specific state yet as all of uses for akita that i've came across have been strictly for sending/retrieving data to display in the app.

client/src/app/employee/employee-create/employee-create.component.html
KyleMcMaster: I'm going to to use `type="submit"` strictly as a note to myself that the button is "submitting" as it's action. Good to know that is the default behavior of `<button>` though.

client/src/app/employee/employee-list/state/employee-list.model.ts
KyleMcMaster: I like this naming convention: `EmployeeListItem`, `Employee`, and `EmployeeDetail`. Updated the list code to use this pattern.

client/src/app/employee/employee-detail/state/employee-detail.store.ts
KyleMcMaster: Updated this `query`, `store`, `service`, and `model` to reflect the naming convention below since this `store` is currently retrieving an `Employee`, I can refactor naming again once we start utilizing the entire `EmployeeDetail` object graph.

client/src/app/employee/employee.component.ts
```diff
@@ -1,20 +1,41 @@
...
+  employee: Observable<Employee>;
+  uiState = '';
+
+  constructor(
+    private employeeQuery: EmployeeQuery,
+    private employeeService: EmployeeService,
+    private employeeListService: EmployeeListService,
+    private employeeListQuery: EmployeeListQuery,
+  ) {
+    this.employee = this.employeeQuery.select((state) => state as Employee);
```
seangwright, line 28: I think you can remove this cast `as Employee`.

Also, if you want to reduce the syntax noise you can move the query assignments up to where the fields are declared:

```typescript
  employees = this.employeeListQuery.selectAll();
  employee = this.employeeQuery.select();
```

client/src/app/employee/employee.component.ts
```diff
@@ -1,20 +1,41 @@
...
   styleUrls: ['./employee.component.scss'],
 })
 export class EmployeeComponent {
-  readonly faLock = faLock;
-  readonly faUnlock = faUnlock;
-  isReadonly = true;
+  readonly faPlus = faPlus;
+  employees: Observable<EmployeeListItem[]>;
+  employee: Observable<Employee>;
+  uiState = '';
```
seangwright, line 20: If you were to change this component to use `OnPush` change detection, you'd want to move this `uiState` into the `EmployeeStore` and react to changes to that state via the query.

This ends up being a pretty typically pattern - move state out of components into classes unrelated to rendering/DOM, then use the component as a binding between DOM and state.

client/src/app/employee/employee.component.ts
```diff
@@ -1,20 +1,41 @@
...
+  }
 
-  constructor() {}
+  setCreate() {
+    this.uiState = 'create';
+  }
 
-  toggleReadonly() {
-    this.isReadonly = !this.isReadonly;
+  setActive(employee: EmployeeListItem) {
```
seangwright, line 37: The Angular style guide recommends naming event handling methods (either from user events like `(click)` or custom component events) with an `on` prefix.

So this method would be named `onSelectEmployee()` and `setCreate()` would become `onCreate()`.

This is helpful from a tooling perspective, beyond just having a convention for naming, because component methods used in the template must be `public` (in Typescript, no access modifier implies `public`), but the Typescript language service doesn't know that the method is referenced in the template so it will display the number of references of the method as 0.

However, by giving the method the `on` prefix, you know that it's used in the template.

Other methods, if they have 0 references, you can safely remove because you know they aren't being used.

client/src/app/employee/employee-list/employee-list.component.ts
```diff
@@ -13,21 +10,20 @@ import { EmployeeListService } from './state/employee-list.service';
...
+  @Input()
+  employees: EmployeeListItem[];
+
+  @Output()
+  selected = new EventEmitter<EmployeeListItem>();
+
+  constructor() { }
 
-  ngOnInit() {}
+  ngOnInit() { }
```
seangwright, line 22: In most presentational components you will find you don't need `ngOnInit()` or `ngOnDestroy()` because there's no special Angular things you are doing with them - they just bind `@Input()` state to the template.

client/src/app/employee/employee-list/employee-list.component.html
```diff
@@ -1,54 +1,22 @@
...
-        'table-secondary': employee.status === 'Disabled'
-      }"
-      *ngFor="let employee of employees | async"
-    >
-      <th scope="row"> {{ employee.firstName + ' ' + employee.lastName }} </th>
-      <td>{{ employee.title }}</td>
-      <td>{{ employee.phone }}</td>
-      <td>{{ employee.employmentStartedOn | date }}</td>
-      <td>
+<div class="row" *ngFor="let employee of employees">
```
seangwright, line 1: There's an argument to be made for making this a Single-File Component (SFC).

There's 0 business logic, the template is small, the component `@Input()` bindings are minimal.

You could even use the `selected` event emitter directly in the template since there's no logic involved:

```html
(click)="selected.emit(employee)"
```

client/src/app/employee/employee-list/state/employee-list.service.ts
```diff
@@ -27,7 +24,7 @@ export class EmployeeListService {
   getEmployees() {
     this.store.setLoading(true);
     return this.http
```
seangwright, line 26: I would remove the `return` here. One thing you can do to prevent yourself from accidentally exposing RxJs stuff from a service is to set the return type on these methods to `void`.

client/src/app/employee/employee-detail/employee-detail.component.html
```diff
@@ -0,0 +1,62 @@
+<div class="row mt-2 mb-2">
```
seangwright, line 1: Another good opportunity for a SFC

client/src/app/employee/employee-detail/state/employee.service.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  getEmployee(id: string) {
+    this.store.setLoading(true);
+
+    this.http
+      .get<Employee>(`${this.apiRootUrl}/Employees/${id}`)
+      .pipe(
+        catchError((err) => {
+          this.toastr.error(`Could not get employee ${id}`);
+          return throwError(err);
```
seangwright, line 31: If the employee request fails, what will currently happen is the error will not be propagated anywhere.

Since the subscription is happening below and there is no `error` callback to the Observer passed to `subscribe()`, the `return throwError(err)` doesn't go anywhere.

It's probably more appropriate to do the `this.toast.error(...)` in the `error` callback of the Observer below.

client/src/app/employee/employee-detail/state/employee.service.ts
```diff
@@ -0,0 +1,68 @@
...
+        }),
+      )
+      .subscribe({
+        next: (detail) => {
+          this.store.update((state) => detail);
```
seangwright, line 36: `update()` also takes a `Partial<T>` where `T` is the state.

This means we can pass `detail` directly:

```typescript
this.store.update(detail);
```

client/src/app/employee/employee-create/employee-create.service.ts
```diff
@@ -3,10 +3,8 @@ import { Injectable } from '@angular/core';
 import { throwError } from 'rxjs';
 import { catchError, tap } from 'rxjs/operators';
 import { EnvService } from 'src/app/shared/env.service';
-import {
-  Employee,
-  EmployeeCreate,
-} from '../employee-list/state/employee-list.model';
+import { Employee } from '../employee-detail/state/employee.model';
+import { EmployeeCreate } from '../employee-list/state/employee-list.model';
 import { EmployeeListStore } from '../employee-list/state/employee-list.store';
```
seangwright, line 8: Try to avoid sharing your store beyond the one service it is associated with. If you start passing around the store to other consumers it becomes harder to predict where state changes are coming from.

Instead, this behavior should probably be in the `employee-list.service` and that service should be used in the `employee-create.component`.

client/src/app/employee/employee.component.ts
KyleMcMaster: fixed

client/src/app/employee/employee.component.ts
KyleMcMaster: Good to know, i like this convention. I need to read the Angular style guide more.

client/src/app/employee/employee-list/employee-list.component.ts
KyleMcMaster: Fixed this throughout multiple components.

client/src/app/employee/employee-detail/state/employee.service.ts
KyleMcMaster: 👍

KyleMcMaster: fixed

client/src/app/employee/employee-create/employee-create.service.ts
KyleMcMaster: removed `employeeCreateService`, added method to `employeeListService`

client/src/app/employee/employee-detail/state/employee.service.ts
KyleMcMaster: fixed, i believe writing all the `error`, `next`, and `complete` code in `subscribe()` means i could get rid of the `pipe()` completely.

seangwright: Yup, in fact the only time you really need `.pipe` is when you don't subscribe.

client/src/app/employee/employee-list/employee-list.component.html
KyleMcMaster: 👍

client/src/app/employee/employee-list/state/employee-list.service.ts
KyleMcMaster: fixed! good recommendation on adding a return type ! 👍

Issue #31, opened 2020-06-13
Consider adding a license


edmistond:
Just for completeness' sake, you might consider applying an appropriate open source license to let people know it's okay to borrow some of these patterns, such as the AutoFixture/NSubstitute setup in PayrollProcessor.Tests.

I generally like MIT license, and it's what the dotnet runtime uses.

KyleMcMaster:
Good idea! @seangwright Thoughts on this?

seangwright:
Yup, good idea. I also use MIT https://opensource.org/licenses/MIT

KyleMcMaster:
Marking issue as resolved after mergin PR #32

PR #32, opened 2020-06-26
Add license and upgrade to Angular v10

KyleMcMaster:
#31 - Adding MIT license to the project

Also soft upgrading all packages for the client as well as upgrade Angular's major version to v10.0


client/tsconfig.base.json
```diff
@@ -0,0 +1,31 @@
+{
+  "compileOnSave": false,
+  "compilerOptions": {
+    "baseUrl": "./",
+    "outDir": "./dist/out-tsc",
+    "sourceMap": true,
+    "declaration": false,
+    "downlevelIteration": true,
+    "experimentalDecorators": true,
+    // TODO - we'll want to enable these incrementally, currently they'll cause compilation errors
```
seangwright, line 10: Heh, looks like they're all enabled now. This will cause build/linting issues.

seangwright:
Did you use https://update.angular.io/?

PR #33, opened 2020-07-10
Payroll List and Department Filter

KyleMcMaster:
Refactor payroll views of the application to use the styling of the employee views. 
Restructure client services to follow practices learned in employee refactor development


client/src/app/employee/employee-detail/state/employee.service.ts
```diff
@@ -1,12 +1,14 @@
...
-import { ToastrService } from 'ngx-toastr';
 import { throwError } from 'rxjs';
 import { catchError } from 'rxjs/operators';
-import { EnvService } from 'src/app/shared/env.service';
+
 import { Employee, EmployeeUpdate } from './employee.model';
 import { EmployeeStore } from './employee.store';
 
+import { ToastrService } from 'ngx-toastr';
+import { EnvService } from 'src/app/shared/env.service';
```
seangwright, line 10: This import path seems wrong - Angular/Typescript should be doing the import from `'./app/shared/env.service'`. I wonder why it sees `src` as the root.

We can set up import aliases so that we'll have absolute imports from `'@payrolls/...'`, `'@shared/...'`

client/src/app/payroll/payroll-list/payroll-list.component.ts
```diff
@@ -1,20 +1,38 @@
...
-export class PayrollListComponent {
+export class PayrollListComponent implements OnInit {
   readonly faSkull = faSkull;
   readonly faSmileBeam = faSmileBeam;
+  readonly payrolls = this.query.selectAll();
+  private employeeDepartment: string;
+
+  @Input()
+  set department(department: string) {
+    const employeeDepartment = department && department.trim();
```
seangwright, line 21: You can rearrange this to get rid of the local variable:

```typescript
this.employeeDepartment = department?.trim() ?? '';

this.service.getPayrolls(employeeDepartment);
```

client/src/app/payroll/payroll-list/payroll-list.component.ts
```diff
@@ -1,20 +1,38 @@
...
+    this.service.getPayrolls(employeeDepartment);
+    this.employeeDepartment = employeeDepartment;
+  }
+
+  get department(): string {
```
seangwright, line 26: I do like the use of getters to simplify values being bound in the template 👍

client/src/app/payroll/payroll-list/payroll-list.component.ts
```diff
@@ -1,20 +1,38 @@
...
-  constructor(private dataService: DataService) {}
+  constructor(
+    private query: PayrollListQuery,
+    private service: PayrollListService,
+  ) {}
 
-  getPayroll(): Payroll[] {
-    return this.dataService.getPayroll();
+  ngOnInit() {
+    this.service.getPayrolls(this.department);
```
seangwright, line 36: The use of an `@Input()` and a `Query` in this component could be an indication of mixed-purposes.

This component could be looking more and more like a presentational component that should accept `@Input()` of employees and not have any service dependencies, which would move the orchestration/data access logic one level up.

client/src/app/payroll/payroll-list/state/payroll-list.service.ts
```diff
@@ -0,0 +1,45 @@
...
+import { ToastrService } from 'ngx-toastr';
+import { EnvService } from 'src/app/shared/env.service';
+import { ListResponse, mapListResponseToData } from 'src/app/shared/list-response';
+
+@Injectable({ providedIn: 'root' })
+export class PayrollListService {
+  readonly apiRootUrl: string;
+
+  constructor(
+    envService: EnvService,
```
seangwright, line 16: Not something you need to (or even should) change, but there's a cute way to get a single value from an injected service/object.

```typescript
constructor(
  { apiRootUrl }: EnvService,
  private readonly http: HttpClient,
  ...
) {
  this.apiRootUrl = apiRootUrl;
}
```

[Playground Link](https://www.staging-typescript.org/play?#code/MYGwhgzhAECC0G8BQ1XQHZgLYFMBc0EALgE4CW6A5tALzQDk9A3EgL5JKiQwBCiKaTLloNmHNNGAB7dMRIBXYESkkAFAImoEGbDmisCsDdACU-TWiIALMhAB0QvXUcsJ7VkA)

client/src/app/payroll/payroll.component.html
```diff
@@ -1 +1,34 @@
...
+      </div>
+      <div
+        class="row"
+        *ngFor="let department of departments"
+        (click)="onSelectDepartment(department)"
+      >
+        <div class="list-group-item list-group-item-action">
+          <div class="row">
+            <div class="col-12">
+              <span>{{ department.replace('_', ' ') }}</span>
```
seangwright, line 23: This would make a great pipe since it's presentational.

```typescript
import { Pipe, PipeTransform } from '@angular/core';

@Pipe({name: 'unslugify'})
export class UnslugifyPipe implements PipeTransform {
  transform(value?: string): string {
    return value?.replace('_', ' ') ?? '';
  }
}
```

Register it in the `SharedModule` declarations and exports, a then use it like:

```html
<span>{{ department | unslugify }}</span>
```

client/src/app/payroll/payroll.component.html
```diff
@@ -1 +1,34 @@
...
+                Departments
+              </span>
+            </div>
+          </div>
+        </div>
+      </div>
+      <div
+        class="row"
+        *ngFor="let department of departments"
+        (click)="onSelectDepartment(department)"
```
seangwright, line 18: Although Angular makes it super easy to add a `(click)` binding to a `<div>`, it's generally a bad idea because `<div>` elements don't include any accessibility features for interactions like clicking/tapping.

If you want a whole 'card' or row to be clickable, then wrapping everything in an `<a href="#">...</a>` and styling it as a button is better.

> The a element can be wrapped around entire paragraphs, lists, tables, and so forth, even entire sections, so long as there is no interactive content within (e.g., buttons or other links). This example shows how this can be used to make an entire advertising block into a link:

[Source](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-a-element)

seangwright: Also, while you're looking at HTML stuff, this is a good post about other HTML5 elements you can use to give your markup more meaning: https://dev.to/kenbellows/stop-using-so-many-divs-an-intro-to-semantic-html-3i9i

client/src/app/payroll/payroll.component.html
KyleMcMaster: Nice! Almost as good as big daddy pipe!

client/src/app/payroll/payroll-list/payroll-list.component.ts
KyleMcMaster: Nice suggestion!

client/src/app/employee/employee-detail/state/employee.service.ts
KyleMcMaster: It looks like this is happening in multiple places for EnvService. How do I set up `@shared` as an import alias?

client/src/app/payroll/payroll-list/payroll-list.component.ts
KyleMcMaster: Refactored to remove `Service` and `Query` and added an input `PayrollListItem[]` from the `PayrollListComponent`. `PayrollComponent` now uses the `Service` and `Query` to get the data and push it into `PayrollListComponent`

client/src/app/employee/employee-detail/state/employee.service.ts
seangwright: https://github.com/microsoft/TypeScript/issues/39666

Something is broken in VS Code or Typescript and it's related to these new "Solution Style" `tsconfig.json` files.

We can switch back to the previous `tsconfig.json` and then import aliases (and `import` intellisense) will work.

seangwright: Looks like the `tsconfig.json` setup that the Angular 10 CLI produces has issues currently and it doesn't appear they will resolved soon.

I switched the structure to use the older `tsconfig.json` setup and added the import path aliases.

KyleMcMaster: Nice! Issues with angular 10 CLI, not good!

client/src/app/payroll/payroll.component.html
KyleMcMaster: Marking as "TODO" since we should come back to this. I have a lot to learn about html and accessibility.

KyleMcMaster:
Merging as discussed with @seangwright . 

Note to come back to this PR for reference on semantic html and accessibility.

PR #35, opened 2020-08-10
Style active department on payrolls view

KyleMcMaster:
Add styling to payrolls feature to highlight selected department in list


client/src/app/payroll/payroll.component.ts
```diff
@@ -14,7 +12,7 @@ export class PayrollComponent {
   readonly faLock = faLock;
   readonly faUnlock = faUnlock;
   readonly payrolls = this.query.selectAll();
-  department = 'Building_Services';
+  selectedDepartment = 'Building_Services';
```
seangwright, line 15: 👍

client/src/styles.scss
```diff
@@ -44,11 +44,15 @@ body {
 }
 
 .background-dark {
-  background-color: #303030 !important;
+  background-color: #303030;
```
seangwright, line 47: Normally we'd want these as SCSS variables, but it doesn't matter too much atm.

PR #36, opened 2020-08-21
Add employee payrolls list

KyleMcMaster:
Add ability to view employee's recent payrolls


client/src/app/app.module.ts
```diff
@@ -18,6 +18,7 @@ import { SharedModule } from '@shared/shared.module';
 
 import { AppRoutingModule } from './app-routing.module';
 import { AppComponent } from './app.component';
+import { EmployeePayrollListComponent } from './employee/employee-payroll-list/employee-payroll-list.component';
```
seangwright, line 21: You should be able to import this from the alias, like the other imports

KyleMcMaster: fixed

client/src/app/employee/employee-list/state/employee-list.service.ts
```diff
@@ -1,17 +1,18 @@
...
+
+import { ToastrService } from 'ngx-toastr';
+
+import { EmployeeDetail } from '@employee/employee-detail/state/employee.model';
+
 import { EmployeeCreate, EmployeeListItem } from './employee-list.model';
 import { EmployeeListStore } from './employee-list.store';
 
+import { EnvService } from 'src/app/shared/env.service';
+import { ListResponse, mapListResponseToData } from 'src/app/shared/list-response';
```
seangwright, line 14: I think these absolute imports were from the solution style `tsconfig.json` that Angular has now reverted from using.

We have the `@shared` alias so these can be changed to `@shared/env.service` and `@shared/list-response`

client/src/app/employee/employee-payroll-list/employee-payroll-list.component.ts
```diff
@@ -0,0 +1,37 @@
+import { Component, Input } from '@angular/core';
+
+import { EmployeePayroll } from '@employee/employee-detail/state/employee.model';
+
+@Component({
+  selector: 'app-employee-payroll-list',
+  template: ` <h5>Recent Payrolls</h5>
```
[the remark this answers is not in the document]
seangwright: Ya, definitely a perfect use case here.

seangwright:
Just 1 item, otherwise looks good to me!

client/src/app/employee/employee-list/state/employee-list.service.ts
KyleMcMaster: fixed these across the app

PR #37, opened 2020-08-31
Update employee details

KyleMcMaster:
- add the ability to update an employees details
- add ability to create employee payrolls 
- update project roadmap to have a verbose checklist
- minor refactorings and cleanup


README.md
```diff
@@ -30,17 +30,29 @@ Areas of interest:
...
 
-- MVP
+### MVP
 
-  - Initial Employee CRUD functionality
-  - Initial Payroll CRUD functionality
+#### Employees
 
-- TODOs
+- [x] Create
```
seangwright, line 37: 👍

client/src/app/employee/employee.component.ts
```diff
@@ -20,8 +20,8 @@ export class EmployeeComponent {
   uiState = '';
 
   constructor(
-    private employeeQuery: EmployeeQuery,
-    private employeeService: EmployeeService,
+    private employeeQuery: EmployeeDetailQuery,
```
seangwright, line 23: I've gotten in the habit of naming constructor dependencies as simply as possible, only increasing the specificity when there is ambiguity.

In this case, because this is the `EmployeeComponent`, I would name the params as follows:

```typescript
private detailQuery: EmployeeDetailQuery,
private detailService: EmployeeDetailService,
private listService: EmployeeListService,
private listQuery: EmployeeListQuery
```

I might even simplify it more and name the primary query and service for the component as `query` and `service`.

Let's say you had a `EmployeeDetailComponent` that needed the `Detail` injectables and the `List` service.

```typescript
private query: EmployeeDetailQuery,
private service: EmployeeDetailService,
private listService: EmployeeListService
```

I've found that by breaking up services and components into smaller units, the shorter names are more readable and differentiating them isn't hard, even though they don't have adjectives/nouns in the param names, because the place where they are being injected provides some context itself.

Of course, this only (potentially) works if the classes are kept small and focused.

Thoughts?

KyleMcMaster: I agree with this convention. It makes sense to remove employee from all of the service/query objects since in the context of these components we are always working with employees. I think the only time we would need the entity in the name is if for example we have a `PayrollDetailService` being injected into an `EmployeeDetailComponent`. I believe in the context of `EmployeeDetailComponent` removing detail from the name also makes sense since we are within the detail context. Adding an `EmployeeListService` as `listService` to the `EmployeeDetailcomponent` would then serve as an example to add more clarity to the name of the service since we would be outside of the scope of `EmployeeListComponent`. We can try to work this convention moving forward and see if any issues come up.

seangwright: Another way to get around the naming issues is to _not_ inject all the services/queries directly into your component.

Instead use the services/queries directly related to your component as the only external dependencies and inject the other services/queries into those.

Example:

```typescript
// employee-detail.component.ts

employee = this.query.employee;
companies = this.query.companies;

constructor(
  private readonly query: EmployeeDetailQuery,
  private readonly service: EmployeeDetailService) { }

onEmployeeDelete(employee: EmployeeDetail) {
  this.service.delete(employee);
}

onCompanyChange(company: Company) {
  this.service.updateCompany(company);
}
```

```typescript
// employee-detail.query.ts
readonly employee = this.select(s => s.employee);
readonly companies = this.companyQuery.selectAll();

constructor(
  private readonly companyQuery: CompanyQuery, 
  private readonly store: EmployeeDetailStore) { super(store); }
```

```typescript
// employee-detail.service.ts
constructor(
  private readonly http: HttpClient,
  private readonly store: EmployeeDetailStore,
  private readonly companyService: CompanyService) { }

delete(employee: EmployeeDetail) {
  this.http.delete('...');
}

updateCompany(company: Company) {
  const { employee } = this.store.getValue();

  this.http.update('...', employee)
    .tap(emp => this.store.update({ employee: emp }))
    .tap(() => this.companyService.refresh());
}
```

client/src/app/employee/employee-create/employee-create.component.ts
```diff
@@ -20,7 +20,14 @@ export class EmployeeCreateComponent {
     title: new FormControl(''),
   });
 
-  departments: Department[] = ['Finance', 'HR', 'IT', 'Sales'];
+  readonly departments: string[] = [
```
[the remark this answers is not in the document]
seangwright: Moving to models seems good enough for now.

If we were to move it to the API, we'd either need to bubble the Observable of its asynchronous access throughout the app (like via a query), or delay the app from loading until the static API data was guaranteed to be pre-loaded into the app (which is a more advanced approach).

client/src/app/employee/employee.component.ts
KyleMcMaster: I decided to do something similar to `updateCompany` in `EmployeeDetailService` by calling the `EmployeeListService` to refresh the list data when employee details are updated.

``` 
 update(employeeUpdate: EmployeeUpdate): void {
    this.store.setLoading(true);

    this.http
      .put<EmployeeDetail>(`${this.apiRootUrl}/Employees`, employeeUpdate)
      .subscribe({
        error: () =>
          this.toastr.error(`Could not update employee ${employeeUpdate.id}`),
        next: (detail) => {
          this.store.update(detail);
          this.listService.getEmployees();
          this.toastr.show('Employee sucessfully updated!');
        },
        complete: () => this.store.setLoading(false),
      });
  }
```

KyleMcMaster (commented):
Fixed comments with suggested changes, leaving private employee detail to use for inputting into child components

KyleMcMaster:
> Do you use this getter? Is the private field `this._employee` still necessary? I think the form has all this info, right?

This

KyleMcMaster:
> Do you use this getter? Is the private field `this._employee` still necessary? I think the form has all this info, right?

Using this to get department, id, and version fields from the employee.

client/src/app/employee/employee.component.ts
seangwright: I saw that - I think that's a nice approach.

api/PayrollProcessor.Web.Api/Features/Employees/EmployeePayrollCreate.cs
```diff
@@ -43,11 +44,13 @@ public class EmployeePayrollCreate : BaseAsyncEndpoint<EmployeePayrollCreateRequ
                 .Dispatch(new EmployeeQuery(request.EmployeeId))
                 .Bind(employee =>
                 {
+                    string payrollPeriod = (ISOWeek.GetWeekOfYear(request.CheckDate.DateTime) / 2).ToString().PadLeft(2, '0');
```
KyleMcMaster, line 47: @seangwright I felt it was more intuitive to let the api determine the payroll period using a check date passed from the client. We currently use the week of the year but in theory there could be an offset if the starting week occurred mid year. This is taken from [this](https://github.com/KyleMcMaster/payroll-processor/blob/master/api/PayrollProcessor.Infrastructure.Seeding/Features/Generators/EmployeeSeed.cs) seed code, is there a shared library we could move this to, or do you think this is fine to duplicate for now?

api/PayrollProcessor.Web.Api/PayrollProcessor.Web.Api.csproj
```diff
@@ -19,6 +19,7 @@
     <PackageReference Include="Swashbuckle.AspNetCore.Annotations" Version="5.4.1" />
     <PackageReference Include="Swashbuckle.AspNetCore" Version="5.4.1" />
     <PackageReference Include="Microsoft.Azure.Cosmos" Version="3.8.0" />
+    <PackageReference Include="Microsoft.AspNetCore.Mvc.NewtonsoftJson" Version="3.1.7"/>
```
KyleMcMaster, line 22: System.Text.Json had issues parsing decimal so I brought in NewtonsoftJson.

reference:
stackoverflow.com/questions/57626878/the-json-value-could-not-be-converted-to-system-int32

KyleMcMaster (commented):
Open to further feedback, getting close to finishing "mvp".

api/PayrollProcessor.Web.Api/PayrollProcessor.Web.Api.csproj
seangwright: Interesting!

api/PayrollProcessor.Web.Api/Features/Employees/EmployeePayrollCreate.cs
seangwright: I say leave it and open an issue to look at converting `EmployeePayrollNew` to a closed domain model with required constructor params.

Then `newPayroll.Period` could be a readonly prop that does the above calculation/transformation.

seangwright (approved):

KyleMcMaster:
Merging, will review any remaining issues inside the EmployeeComponent tree in the next PR.

PR #41, opened 2020-09-11
Fix/client forms package updates

seangwright:
# Description

- Client fix for Employee Payroll creation
- Client package updates to latest
- Client add new `clock.service` to remove `new Date()` dependency

## Purpose

- [X] Feature
- [ ] Partial Feature
- [X] Bugfix
- [ ] Docs, Config
- [ ] Chore / Format

## Change Type

- [ ] Major
- [ ] Minor
- [X] Fix

## Data Persistence Changes

n/a

## Configuration Changes

I noticed issues when running the client after installing the package updates. Lots of warnings about sourcemaps.

I removed `node_modules` and re-ran the `Client: Install` task and everything worked correctly.


client/src/app/employee/employee-payroll-create/employee-payroll-create.component.ts
```diff
@@ -1,33 +1,35 @@
...
     this.filterForm.patchValue({
       employeeId: employee.id,
+      checkDate: this.clock.now(),
+      grossPayroll: 0,
     });
   }
 
-  constructor(private detailService: EmployeeDetailService) {}
-
-  ngOnChanges(changes: SimpleChanges): void {
```
seangwright, line 28: This was causing the form values to be cleared out because the `@Input()` binding is updated first, then `OnChanges()` runs.

So the `employeeId` value was assigned to the form but then the form was reset and it lost the value, which caused an API request validation error.

client/src/app/shared/clock.service.ts
```diff
@@ -0,0 +1,10 @@
+import { Injectable } from '@angular/core';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class ClockService {
+  now(): Date {
```
seangwright, line 7: A simple abstraction to uncouple the dependency on the system clock with `new Date()`

KyleMcMaster (approved):
LGTM, thanks for upgrading the packages.

PR #43, opened 2020-09-11
Feat/client lazy routes module cleanup

seangwright:
# Description

- Migrates all modules / routes to lazy loaded
- Adds `Core` module for future core app functionality (currently only has catch-all route and `NotFound` component)
- Adds `Admin` module for future administration features (data refresh request to Functions API or getting stats from the cosmos db / storage queue)
- Replaces nav header ng bootstrap directives with router integration

## Purpose

- [X] Feature
- [X] Partial Feature
- [ ] Bugfix
- [ ] Docs, Config
- [ ] Chore / Format

## Change Type

- [X] Major
- [ ] Minor
- [ ] Fix

## Data Persistence Changes

n/a

## Configuration Changes

n/a


client/src/app/app-routing.module.ts
```diff
@@ -1,9 +1,24 @@
...
 import { RouterModule, Routes } from '@angular/router';
 
-import { EmployeeComponent } from './employee/employee.component';
-
-const routes: Routes = [{ path: 'employees', component: EmployeeComponent }];
+const routes: Routes = [
+  {
+    path: 'employees',
+    loadChildren: () =>
+      import('./employee/employee.module').then((m) => m.EmployeeModule),
```
seangwright, line 8: This syntax ensures that the token `EmployeeModule` is not discovered by the compiler as being a dependency of this ^ (`App.Module.ts`) module, which means the webpack bundler can separate out the entire contents of `EmployeeModule`, and everything it contains, into a separate file/chunk.

This results in the initial app payload not including any `EmployeeModule` stuff and it's only downloaded when the user routes to `/employees`.

This applies as well to the other lazy loaded modules in this route definition.

https://ultimatecourses.com/blog/lazy-load-angular-modules

client/src/app/app.component.html
```diff
@@ -1,25 +1,15 @@
...
-            </ng-template>
+        <ul ngbNav [activeId]="fragment | async" class="navbar-nav">
+          <li [ngbNavItem]="link.path" *ngFor="let link of links">
+            <a ngbNavLink [routerLink]="[link.path]">{{ link.title }}</a>
           </li>
         </ul>
       </div>
     </nav>
-    <div [ngbNavOutlet]="nav"></div>
     <router-outlet></router-outlet>
```
seangwright, line 13: The components will now be displayed by the `<router-outlet>` instead of the `[ngbNavOutlet]`

client/src/app/app.module.ts
```diff
@@ -1,52 +1,35 @@
...
 import { ToastrModule } from 'ngx-toastr';
+
+import { SharedModule } from '@shared/shared.module';
+
 import { AppRoutingModule } from './app-routing.module';
 import { AppComponent } from './app.component';
-import { EmployeePayrollCreateComponent } from './employee/employee-payroll-create/employee-payroll-create.component';
+import { CoreModule } from './core/core.module';
+
+const NG_MODULES = [BrowserAnimationsModule, BrowserModule, HttpClientModule];
```
seangwright, line 16: This is a pattern I like to use to help organize module imports, declarations, and exports.

client/src/app/employee/employee.module.ts
```diff
@@ -0,0 +1,38 @@
...
+    EmployeeComponent,
+    EmployeeCreateComponent,
+    EmployeeDetailComponent,
+    EmployeeListComponent,
+    EmployeePayrollCreateComponent,
+    EmployeePayrollListComponent,
+  ],
+  imports: [
+    RouterModule.forChild(routes),
+    NgbModule,
```
seangwright, line 31: Eventually, we can replace this `NgbModule` import with explicit declarations of the components we use. It's better to bring in only whats needed because that makes it a lot easier to remove stuff later.

client/src/app/app.component.html
KyleMcMaster: Interesting, I'm not really familiar with angular routing constructs yet so i'll have to look up more about this.

client/src/app/app-routing.module.ts
KyleMcMaster: Good to know, cool to see this implemented so I can follow how this all works. Interesting to see the changes you made in `app-routing.module`  and how that lines up with the `links` array in the `appcomponent`.

KyleMcMaster:
LGTM

KyleMcMaster (approved):

client/src/app/app-routing.module.ts
seangwright: Yes, the route / path management can be centralized. There is even a library to help give strong typing to routes.

PR #47, opened 2020-09-16
Refactor: ICommandHandler return type to `TryAsync`

### LINKED ISSUE #46: Suggestion: Change `ICommandHandler` type to return `TryAsync` instead of `TryOptionAsync`
The `TryOptionAsync` type is good at representing an async operation that can result in Success/Fail and in the Success case there is a Some/None value.

This makes sense for our queries because the query itself could succeed or fail (throw exception) and the query itself could return the item(s) or nothing, if they were not found.

However, commands typically don't return anything (except maybe an identifier from a completed operation) and they aren't necessarily looking for anything in particular. If an item that is needed for a command can't be found, that command is probably best considered to be failed.

In summary, we should consider changing `ICommandHandler<TCommand>` to have a `Execute()` method with the signature

```csharp
TryAsync<Unit> Execute(TCommand command, CancellationToken token);
```

seangwright:
## Description

Fixes #46 and makes command handlers more functional using `LanguageExt` features.

## Purpose

- [X] Feature
- [ ] Partial Feature
- [ ] Bugfix
- [ ] Docs, Config
- [ ] Chore / Format

## Change Type

- [X] Major
- [ ] Minor
- [ ] Fix

## Data Persistence Changes

n/a

## Configuration Changes

n/a


api/PayrollProcessor.Data.Persistence/Features/Departments/DepartmentEmployeeUpdateCommandHandler.cs
```diff
@@ -20,46 +22,55 @@ public DepartmentEmployeeUpdateCommandHandler(CosmosClient client)
...
+                        .Apply(TryAsync),
+                    (record, iterator) => new { record, iterator }
+                )
+                .MapAsync(async aggregate =>
                 {
-                    foreach (var result in await iterator.ReadNextAsync(token))
+                    var record = aggregate.record;
+                    var iterator = aggregate.iterator;
+
+                    while (iterator.HasMoreResults)
```
seangwright, line 57: No clue 🤷‍♂️ what the best techniques are for converting a `foreach` nested in a `while` to use `LanguageExt` types.

api/PayrollProcessor.Data.Persistence/Features/Employees/EmployeePayrollUpdateCommandHandler.cs
```diff
@@ -24,25 +25,24 @@ public EmployeePayrollUpdateCommandHandler(CosmosClient client, IQueueClientFact
...
         }
 
-        public TryOptionAsync<EmployeePayroll> Execute(EmployeePayrollUpdateCommand command, CancellationToken token) =>
-            async () =>
-            {
-                var record = EmployeePayrollRecord.Map.Merge(command); ;
-
-                var updateResponse = await client
+        public TryAsync<EmployeePayroll> Execute(EmployeePayrollUpdateCommand command, CancellationToken token) =>
+            command.Apply(EmployeePayrollRecord.Map.Merge)
```
seangwright, line 29: Not sure which style I prefer. This could also be written as

```csharp
EmployeePayrollRecord.Map.Merge(command)
  .Apply(record => client...)
```

api/PayrollProcessor.Data.Persistence/Infrastructure/Clients/CosmosClientExtensions.cs
```diff
@@ -30,4 +30,9 @@ public static class CosmosClientExtensions
                 PartitionKey = new PartitionKey(partitionKey.ToLowerInvariant())
             });
     }
+
+    public static class CosmosResponse
+    {
+        public static T Unwrap<T>(Response<T> response) => response.Resource;
```
seangwright, line 36: In the future we might want to validate the `StatusCode` property of the `Response<T>`... right now we assume everything is 👍

api/PayrollProcessor.Functions.Api/Features/Departments/DepartmentEmployeeTrigger.cs
```diff
@@ -77,7 +78,10 @@ public class DepartmentEmployeeTrigger
                     .Bind(
                         employee => commandDispatcher.Dispatch(new DepartmentEmployeeCreateCommand(employee, idGenerator.Generate()))
                             .DoIfFail(ex => log.LogError(ex, "Could not create Department Employee for {@employee}", employee))
-                            .Do(de => log.LogInformation("{@departmentEmployee} created for {employeeId}", de, message.EmployeeId))))
+                            .Do(de => log.LogInformation("{@departmentEmployee} created for {employeeId}", de, message.EmployeeId))
+                            .ToTryOption()
+                    )
```
seangwright, line 83: Not sure if having the parens match with the method calls in indentation is better or if having them all line up at the end of the last call is better (see below)

api/PayrollProcessor.Web.Api/Features/Employees/EmployeeUpdate.cs
```diff
@@ -33,23 +33,20 @@ public EmployeeUpdate(IQueryDispatcher queryDispatcher, ICommandDispatcher comma
...
-                        request.Title,
-                        request.Version,
-                        employee
-                    );
-
-                    return commandDispatcher.Dispatch(command);
-                })
+            queryDispatcher
+                .Dispatch(new EmployeeQuery(request.Id), token)
+                .Map(employee => new EmployeeUpdateCommand(
```
seangwright, line 38: Convert this to an expression to get rid of the statement/assignment.

Here's an example of where a static factory method on `EmployeeUpdateCommand` would be nice because we could call it via `.Map(EmployeeUpdateCommand.Create)` ... however, it would have to take `Employee` as a parameter, not the individual primitives.

This might be an ok approach since they are both `Core.Domain` types.

KyleMcMaster (approved):
LGTM, lots to unpack here.

PR #52, opened 2021-02-21
Update angular packages to v11

KyleMcMaster:
- update angular packages to v11

## Purpose

- [ ] Feature
- [ ] Partial Feature
- [ ] Bugfix
- [ ] Docs, Config
- [x] Chore / Format

## Change Type

- [ ] Major
- [x] Minor
- [ ] Fix

## Data Persistence Changes

N/A

## Configuration Changes

N/A


client/angular.json
```diff
@@ -46,7 +46,6 @@
               "optimization": true,
               "outputHashing": "all",
               "sourceMap": false,
-              "extractCss": true,
```
seangwright, line 49: Is this no longer a feature in ng11? Extracting the CSS can be nice because CSS can be parsed in parallel w/ the javascript, whereas, with embedded CSS, the javascript needs parsed and executed before the CSS is inserted into the `<head>`

KyleMcMaster: this was removed by default when running `ng update`.
image

seangwright (approved):
👍

We'll want to look at moving off tslint. But that's a bigger task.

PR #53, opened 2021-03-14
Implement basic auth flow with Auth0

KyleMcMaster:
# Description

Implements a basic authentication flow using auth0 as the identity provider. 

TODO: update documentation for setting up dev's own auth0 account

## Purpose

- [ ] Feature
- [x] Partial Feature
- [ ] Bugfix
- [ ] Docs, Config
- [ ] Chore / Format

## Change Type

- [x] Major
- [ ] Minor
- [ ] Fix

## Data Persistence Changes

N/A

## Configuration Changes

app.module needs configuration values for :
`
    domain: '',
    clientId: '',
`


api/PayrollProcessor.Core.Domain/Features/UserSessions/UserSession.cs
```diff
@@ -0,0 +1,22 @@
+using System;
+
+namespace PayrollProcessor.Core.Domain.Features.UserSessions
+{
+    public class UserSession
```
seangwright, line 5: Wondering now if these would be better served by C#9 records?

api/PayrollProcessor.Core.Domain/Features/Users/UserCreateCommand.cs
```diff
@@ -0,0 +1,21 @@
...
+namespace PayrollProcessor.Core.Domain.Features.Users
+{
+    public class UserCreateCommand : ICommand<User>
+    {
+        public Guid NewId { get; set; }
+        public UserNew User { get; }
+
+        public UserCreateCommand(Guid newId, UserNew user)
+        {
+            Guard.Against.Default(newId, nameof(newId));
```
seangwright, line 14: I was just doing some research into updates to FluentValidation

I think that creating validators for Queries/Commands and running the Validation as the outer most decorator for the Handlers would be a better approach to Guarding so that we could return failed Results instead of throwing exceptions.

api/PayrollProcessor.Data.Persistence/Features/Users/UserCreateCommandHandler.cs
```diff
@@ -0,0 +1,36 @@
...
+
+        public UserCreateCommandHandler(CosmosClient client)
+        {
+            Guard.Against.Null(client, nameof(client));
+
+            this.client = client;
+        }
+
+        public TryAsync<User> Execute(UserCreateCommand command, CancellationToken token) =>
+            UserRecord
```
seangwright: Methods like these has me missing LanguageExt.

client/src/app/access/login/login.component.ts
```diff
@@ -0,0 +1,17 @@
...
+@Component({
+  selector: 'app-login',
+  template: `
+    <button (click)=loginWithRedirect() class="btn-dark">Log in</button>
+  `,
+})
+export class LoginComponent {
+  constructor(private authService: AuthService) {}
+
+  loginWithRedirect() {
```
seangwright, line 14: The naming convention I tend to follow with Angular event handlers is using an `on` prefix.

That way, when viewing the list of methods in a component you know which are event handlers very quickly.

So, I would name this `onLoginClick`.

client/src/app/access/login/login.component.ts
```diff
@@ -0,0 +1,17 @@
+import { Component } from '@angular/core';
+
+import { AuthService } from '@auth0/auth0-angular';
+
+@Component({
+  selector: 'app-login',
+  template: `
+    <button (click)=loginWithRedirect() class="btn-dark">Log in</button>
```
seangwright, line 8: Definitely a good use case for SFCs.

client/src/app/access/logout/logout.component.ts
```diff
@@ -0,0 +1,24 @@
+import { DOCUMENT } from '@angular/common';
```
seangwright, line 1: When was this added 🤔?

I always used to create my own wrappers around `window` and `document` globals.

client/src/app/app.module.ts
```diff
@@ -7,15 +7,22 @@ import { FontAwesomeModule } from '@fortawesome/angular-fontawesome';
...
+import { LogoutComponent } from './access/logout/logout.component';
 import { AppRoutingModule } from './app-routing.module';
 import { AppComponent } from './app.component';
 import { CoreModule } from './core/core.module';
 
 const NG_MODULES = [BrowserAnimationsModule, BrowserModule, HttpClientModule];
 
 const THIRD_PARTY_MODULES = [
+  AuthModule.forRoot({
+    domain: 'DOMAIN',
```
seangwright, line 23: Should these come from 'app settings'/'environment.ts'?

client/src/app/shared/authentication/require-authentication-guard.ts
```diff
@@ -0,0 +1,50 @@
...
+  canActivate(
+    route: ActivatedRouteSnapshot,
+    state: RouterStateSnapshot,
+  ):
+    | boolean
+    | UrlTree
+    | Observable<boolean | UrlTree>
+    | Promise<boolean | UrlTree> {
+
+    return this.authService.isAuthenticated$;
```
seangwright, line 33: 👍

client/src/app/shared/authentication/require-unauthenticated-guard.ts
```diff
@@ -0,0 +1,46 @@
...
+  canLoad(
+    route: Route,
+    segments: UrlSegment[],
+  ):
+    | boolean
+    | UrlTree
+    | Observable<boolean | UrlTree>
+    | Promise<boolean | UrlTree> {
+    return this.authService.isAuthenticated$.pipe(
+      map((b) => !b),
```
seangwright, line 43: I think, generally, you don't want to allow loading of a module (downloading it w/ async routes), if a user won't be able to activate it.

Was there an issue with this method having the same logic as `canActivate`?

client/src/app/access/logout/logout.component.ts
KyleMcMaster: Not sure! Most of this copde / functionality is from https://auth0.github.io/auth0-angular/index.html#getting-started

client/src/app/shared/user-session/state/user-session.service.ts
```diff
@@ -0,0 +1,33 @@
...
+  constructor(
+    envService: EnvService,
+    private http: HttpClient,
+    private store: UserSessionStore,
+    private toastr: ToastrService,
+  ) {
+    this.apiRootUrl = envService.apiRootUrl;
+  }
+
+  getUserSession(token: string) {
```
seangwright: I think the commented out section below is actually what you want. This method should return `: void`

client/src/app/app-routing.module.ts
```diff
@@ -1,19 +1,30 @@
...
+  RouterModule,
+  Routes,
+} from '@angular/router';
+
+import { RequireAuthenticatedGuard } from '@shared/authentication/require-authentication-guard';
 
 const routes: Routes = [
   {
     path: 'employees',
+    canActivate: [RequireAuthenticatedGuard],
```
seangwright, line 12: I was never very happy with all this repetition of guards when an entire app required authentication with the exception of login/forgot-password.

The best way to solve it that I came up with was putting all of the auth'd routes within their own module via a namespace in the URL, like `/a/`.

I think it might also be possible with a separate routing module with an empty route path `''`, but I'd have to test and see if there were issues with route resolution.

KyleMcMaster:
abandoning to simplify dev experience. might pursue this in some future state when other enhancements have been completed.

PR #141, opened 2023-09-10
Convert Employee Commands to CSharpFunctionalExtensions

KyleMcMaster:
# Description

- Adds foundation for migrating to CSharpFunctionalExtensions feature by feature rather than an all or nothing "big bang" approach by using the Strangler Fig pattern. 

## Purpose

- [ ] Feature
- [x] Partial Feature
- [ ] Bugfix
- [ ] Docs, Config
- [ ] Chore / Format

## Change Type

- [x] Major
- [ ] Minor
- [ ] Fix

## Data Persistence Changes

N/A

## Configuration Changes

N/A

### DIFF
--- a/api/PayrollProcessor.Core.Domain/Intrastructure/Operations/Commands/StranglerCommandDispatcher.cs
+++ b/api/PayrollProcessor.Core.Domain/Intrastructure/Operations/Commands/StranglerCommandDispatcher.cs
@@ -1,0 +1,84 @@
+using System;
+using System.Collections.Concurrent;
+using System.Threading;
+using Ardalis.GuardClauses;
+using CSharpFunctionalExtensions;
+using PayrollProcessor.Core.Domain.Intrastructure.Operations.Factories;
+
+namespace PayrollProcessor.Core.Domain.Intrastructure.Operations.Commands;
+
+/// <summary>
+/// TODO: Temporarily named after the Strangler Fig Pattern as this serves as an implementation of <see cref="ICommandDispatcher"/> migrating to CSharpFunctionalExtensions from LanguageExt.
+/// </summary>
+public class StranglerCommandDispatcher : IStranglerCommandDispatcher
+{
+    private readonly ServiceProviderDelegate serviceProvider;
+    private static readonly ConcurrentDictionary<Type, object> commandHandlers = new ConcurrentDictionary<Type, object>();
+
+    public StranglerCommandDispatcher(ServiceProviderDelegate serviceProvider)
+    {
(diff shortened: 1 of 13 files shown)


api/PayrollProcessor.Core.Domain/Intrastructure/Operations/Commands/StranglerCommandDispatcher.cs
```diff
@@ -0,0 +1,84 @@
...
+using System.Collections.Concurrent;
+using System.Threading;
+using Ardalis.GuardClauses;
+using CSharpFunctionalExtensions;
+using PayrollProcessor.Core.Domain.Intrastructure.Operations.Factories;
+
+namespace PayrollProcessor.Core.Domain.Intrastructure.Operations.Commands;
+
+/// <summary>
+/// TODO: Temporarily named after the Strangler Fig Pattern as this serves as an implementation of <see cref="ICommandDispatcher"/> migrating to CSharpFunctionalExtensions from LanguageExt.
```
seangwright, line 11: Have you done this kind of naming in a project before? I haven't but I like the idea of calling this type out explicitly. That way, you know it should eventually be renamed when it takes over.

api/PayrollProcessor.Web.Api/Features/Employees/EmployeeCreate.cs
```diff
@@ -34,7 +35,7 @@ public EmployeeCreate(ICommandDispatcher dispatcher, IEntityIdGenerator generato
         OperationId = "Employees.Create",
         Tags = new[] { "Employees" })
     ]
-    public override Task<ActionResult<Employee>> HandleAsync([FromBody] EmployeeCreateRequest request, CancellationToken token)
+    public override ActionResult<Employee> Handle([FromBody] EmployeeCreateRequest request)
```
seangwright, line 38: Do you drop the `Async` suffix most of the time now? I think Microsoft s current guidance is to only use it if there is also an existing sync API of the same name...

But I've seen a lot of modern code go both ways.

api/PayrollProcessor.Core.Domain/Intrastructure/Operations/Commands/StranglerCommandDispatcher.cs
KyleMcMaster: I haven't done this before but the approach was mentioned in Code That Fits In Your Head and I couldn't think of a better name so I figured calling it out explicitly would be best while the code is in migration state.

